PR #58, opened 2014-12-26
Refactor quattor/schema

stdweird:
`quattor/schema` can't be used from the `template-library-core` repository due to the missing `monitoring/lemon/schema`. 
With the refactor,it is now possible for components to include only the component relevant types and nothing else, so we can unittest the templates shipped by e.g. the components with only dependency on this repo.


stdweird:
If https://github.com/quattor/maven-tools/pull/42 gets merged/released, this PR will also be required, because component schemas can't be validated without it.
@jouvin any further remarks/objections?

jrha:
Can you update the annotations to use `desc = blah`, then I think this is good to go.

quattor/blockdevices.pan
```diff
@@ -1,41 +1,27 @@
...
-# type definition for block devices.
-#
-#
-#
-# DESCRIPTION: @DESCRIPTION@
-
-#
-############################################################
+@{
+    Type definitions for block devices.
```
jrha:
```pan
desc = Type definitions for block devices
```

quattor/blockdevices.pan
```diff
@@ -61,19 +46,19 @@ type blockdevices_partition_type = {
 	"flags" ? blockdevices_partition_flags
 };
 
-#
-# Software RAID using the MD device
-#
+@{
+    Software RAID using the MD device
```
jrha: `desc = Software RAID using the MD device`

etc.

quattor/blockdevices.pan
stdweird: reading http://quattor-pan.readthedocs.org/en/latest/pan-book/pan-book.html?highlight=annotation, `desc` seems to be used for functions (and require the `@documentation{}`).

stdweird:
@jrha i updated according to panbook.

jrha:
Nice, looks good now, @jouvin please merge if you are happy.

PR #60, opened 2015-03-01
Update panc compiler version to use for Quattor development

jouvin:


jouvin:
I suggest adding to 15.2 if possible...

jrha:
Hmmmmm, ok.

PR #62, opened 2015-03-15
README: update the URL about template library update

jouvin:


README.md
```diff
@@ -8,7 +8,7 @@ This repository is intended to be used in conjunction with other parts of the te
 library (standard, os, grid...) to actually configure a system. To help putting together
 everything needed in a way directly usable in a real configuration, it is recommended to
 follow the procedure describe at 
-https://trac.lal.in2p3.fr/Quattor/wiki/Download/QWGTemplates/Install#Upgradinganexistingrepository.
+http://www.quattor.org/documentation/2014/06/06/how-to-use-template-library.html#how-to-get-it?.
```
ned21: Is the ? at the end of the URL necessary? it looks a bit odd. I think it would also look a bit better if the text was: "it is recommended to follow this `[procedure](...link text...)".`

jouvin: I'll update the text as suggested, clearly better! the `? ` is unfortunately part of the URL... This is coming from the anchor automatically generated by Jekyl, because the title ends with a `? `. This mistake was overlooked initially, If you think this is better, I can update the referred page to remove the `?` in the title.

ned21: ah - makes sense. The ? is not treated as part of the link so was appearing verbatim. Using the `[...](..)` syntax will fix that problem too.

jouvin: I'll do it, thanks for the suggestion. I originally didn't do it because I wanted to keep this file as a minimalistic markdown file for people reading it as a text file. But in this particular case, you are right, it will make the things clearer.

jouvin:
Please, keep this PR open until further notice. I use it to test my changes to `get-template-library`...

jouvin:
This PR can be merged after https://github.com/quattor/release/pull/79.

PR #74, opened 2015-08-13
Deprecate shipping of component code in profiles

jrha:
Related to quattor/configuration-modules-core#553

### DIFF
--- a/quattor/types/component.pan
+++ b/quattor/types/component.pan
@@ -18,6 +18,6 @@ type structure_component = extensible {
     "version"           ? string with match(SELF, '^\d+\.\d+\.\d+$')
     "register_change"   ? element*[]
     "dependencies"      ? structure_component_dependency
-    "code"              ? structure_component_code
+    "code"              ? structure_component_code with { deprecated(0, "The code property has been deprecated, support for shipping component code in profiles was removed in 15.2"); true; }
     "ncm-module"        ? string with match(SELF, '^([a-zA-Z_]\w*(?:::[a-zA-Z_]\w*)*)$')
 };


quattor/types/component.pan
```diff
@@ -18,6 +18,6 @@ type structure_component = extensible {
     "version"           ? string with match(SELF, '^\d+\.\d+\.\d+$')
     "register_change"   ? element*[]
     "dependencies"      ? structure_component_dependency
-    "code"              ? structure_component_code
+    "code"              ? structure_component_code with deprecated(0, "Shipping component code in the profiles is deprecated and is non-functional.")
```
stdweird: should return `true` like 
https://github.com/quattor/configuration-modules-core/blob/master/ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan#L18

quattor/types/component.pan
```diff
@@ -18,6 +18,6 @@ type structure_component = extensible {
     "version"           ? string with match(SELF, '^\d+\.\d+\.\d+$')
     "register_change"   ? element*[]
     "dependencies"      ? structure_component_dependency
-    "code"              ? structure_component_code
+    "code"              ? structure_component_code with { deprecated(0, "daemon property has been deprecated, daemons should be used instead"); true; }
```
ned21: s/daemon/code/

jrha:
Unit tests will fail until other problems in this repo are fixed and not before RC1.

stdweird:
@jrha don't know if we agreed to anything wrt that yesterday: should we already merge or wait for rc1 without template-library-core?

stdweird:
hmm, but you merged #77. so i guess it's ok

jrha:
Don't wait, just merge.

PR #75, opened 2015-08-13
Remove component code parameter

jrha:
Based on and finishes #74

### DIFF
--- a/quattor/types/component.pan
+++ b/quattor/types/component.pan
@@ -7,17 +7,11 @@ type structure_component_dependency = {
     "post"      ? string[] with is_component_list(SELF)
 };
 
-type structure_component_code = {
-    "script"    : string
-    "data"      ? string{}
-};
-
 type structure_component = extensible {
     "active"            : boolean = true # "component active?"
     "dispatch"          : boolean = true # "run with cdispd?"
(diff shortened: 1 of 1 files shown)


ned21:
First commit needs to be removed since it is made redundant by the second. If this functionality was removed some time back then I think it should be OK just to yank the structure without a deprecation warning first. People may need to do some code clean up as part of their upgrade, but that's OK - they will get a hard error, and there's an easy local workaround if they really can't delete that code.

jrha:
I discussed this with @stdweird and we settled on the two stage approach, as it is possible that sites are still writing to the parameter (useless but harmless), so just yanking it seems mean.

stdweird:
retest this please

stdweird:
@jrha the tests... they are... :green_apple: :green_book: :green_heart: :tea: green
might want to turn this in a protected repo

jrha:
!!!!!
:astonished:

Done.

jrha:
Are we expecting anything else from the tests than this?
> pan annotations ok
>
> [OK] expected panc failure due to missing data.

stdweird:
@jrha euhm, i can't merge things, but LGTM
@ned21 can you press shiny green button?

PR #76, opened 2015-08-17
is_fqdn: only match FULLY QUALIFIED domain names

### LINKED ISSUE #66: is_fqdn function also matches partially qualified hostnames
The [`is_fqdn`](https://github.com/quattor/template-library-core/blob/master/pan/types.pan#L431) function in [types.pan](https://github.com/quattor/template-library-core/blob/master/pan/types.pan) will match partially qualified hostnames as well as proper FQDNs.
For instance, it will match `example.org` (which it should) but will also match `example`.
The [regex](https://github.com/quattor/template-library-core/blob/master/pan/types.pan#L436) will match a properly formed hostname component followed by a literal `.` or the end of the string.

This means there is no functionality available to ensure that a value is actually an FQDN as opposed to a hostname.

Furthermore, the [`is_hostname`](https://github.com/quattor/template-library-core/blob/master/pan/types.pan#L448) function seems to [rely](https://github.com/quattor/template-library-core/blob/master/pan/types.pan#L453) on this behaviour (it should match IPs, FQDNs and partially qualified hostnames) and would be (partially) broken by a fix to `is_fqdn`.

jrha:
Fixes #66

Tested and seems to work as desired!

### DIFF
--- a/pan/types.pan
+++ b/pan/types.pan
@@ -829,6 +829,19 @@ type type_uppercase = string with {
 
 
 @documentation{
+Checks if the argument is a valid top level domain name (e.g. .example)
+}
+function is_top_level_domain = {
+    match(ARGV[0], '^(\.)+[a-zA-Z]{2,63}?$');
+};
+
+
+type type_top_level_domain = string with {
+    is_top_level_domain(SELF);
+};
+
+
+@documentation{
 Checks if the argument is in the form host.name.domain or IP,
 or .domain or IP/mask.
 }
@@ -837,9 +850,9 @@ function is_network_name = {
         error ("usage: is_network_name (string)");
     };
 
-    if (is_hostname (ARGV[0])) {
-        return (true);
-    };
+    if (is_hostname(ARGV[0])) return(true);
+
+    if (is_top_level_domain(ARGV[0])) return(true);
 
     # Not a hostname. Is it a IP/mask?
     hst = ARGV[0];
@@ -847,14 +860,10 @@ function is_network_name = {
     ip = substr (hst, 0, pos);
     mask = substr (hst, pos+1);
     if (is_ip (ip) && (is_ip (mask) || match (mask, '^(\d+)$'))) {
-        return (true);
-    };
-    # Not IP/mask: is it .domain?
-    if (index ('.', hst) == 0) {
-        return (is_hostname (substr (hst, 1)));
+        return(true);
     };
     # Everything failed!
-    return (false);
+    return(false);
 };
 
 
--- a/pan/types.pan
+++ b/pan/types.pan
@@ -422,7 +422,7 @@ type type_ip = string with {
 @documentation{
 Function to validate a fully-qualified domain name.  Each part
 of the domain name is separated by a period.  The individual parts
-must begin with a letter or digit, end with a letter or digit, 
+must begin with a letter or digit, end with a letter or digit,
 and may contain letters, digits, or hyphens in the middle.
 
 The relevant RFC's for host name syntax are 952, 1053, and 1123
--- a/pan/types.pan
+++ b/pan/types.pan
@@ -841,6 +841,22 @@ type type_top_level_domain = string with {
 };
 
 
+function is_ipv4_netmask_pair = {
+    hst = ARGV[0];
+    pos = index ("/", hst);
+    ip = substr (hst, 0, pos);
+    mask = substr (hst, pos+1);
+    if (is_ip (ip) && (is_ip (mask) || match (mask, '^(\d+)$'))) {
+        return(true);
+    };
+};
+
+
+type type_ipv4_netmask_pair = string with {
+    is_ipv4_netmask_pair(SELF);
+};
+
+
 @documentation{
 Checks if the argument is in the form host.name.domain or IP,
 or .domain or IP/mask.
@@ -860,13 +876,8 @@ function is_network_name = {
     if (is_top_level_domain(ARGV[0])) return(true);
 
     # Not a hostname. Is it a IP/mask?
-    hst = ARGV[0];
-    pos = index ("/", hst);
-    ip = substr (hst, 0, pos);
-    mask = substr (hst, pos+1);
-    if (is_ip (ip) && (is_ip (mask) || match (mask, '^(\d+)$'))) {
-        return(true);
-    };
+    is_ipv4_netmask_pair(ARGV[0]);
+
     # Everything failed!
     return(false);
 };
--- a/pan/types.pan
+++ b/pan/types.pan
@@ -842,13 +842,15 @@ type type_top_level_domain = string with {
 
 
 function is_ipv4_netmask_pair = {
-    hst = ARGV[0];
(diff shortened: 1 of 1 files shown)


pan/types.pan
```diff
@@ -433,7 +433,7 @@ function is_fqdn = {
     if (ARGC != 1 || !is_string(ARGV[0]))
         error("usage: is_fqdn(string)");
 
-    return(match(ARGV[0],'^([a-zA-Z\d]([a-zA-Z\d-]{0,253}[a-zA-Z\d])?(\.|$))+$'));
+    return(match(ARGV[0],'(?=^.{4,253}$)(^((?!-)[a-zA-Z0-9-]{1,63}(?<!-)\.)+[a-zA-Z]{2,63}\.?$)'));
```
stdweird: inspired by http://stackoverflow.com/questions/11809631/fully-qualified-domain-name-validation?

stdweird:
LGTM

pan/types.pan
jrha: Looks the same @gvasilak didn't cite a source.

jrha:
NB: This will probably break `is_hostname` as it stands.

stdweird:
@jrha why would it do that?

jrha:
Hmm, @gvasilak seemed to think it would, but it would actually cause it to behave as documented so...

NB: This also fixes `is_hostname`!

stdweird:
and breaks `is_network_name`, since it checks if `.domain` via `is_hostname(domain)`, good luck verifying TLDs.

stdweird:
also, now that shorthostnames aren't valid hostnames anymore, `is_network_name` also fails on shorthostnames, which is an issue for our ceph config.

jrha:
`is_network_name` now matches TLDs again, according to it's "documentation" it shouldn't match short host names. So do we change the `ncm-ceph` schema or `is_network_name`?

stdweird:
can we issue a deprecated warning to allow short hostnames as part of the backwards incompatibility? and clean it up to conformto the docs in next release? (or is this not worth it?)

jrha:
I guess we can.

pan/types.pan
```diff
@@ -829,6 +829,20 @@ type type_uppercase = string with {
 
 
 @documentation{
+Checks if the argument is a valid top level domain name (e.g. .example)
+}
+function is_top_level_domain = {
+    if (match(ARGV[0], '^(\.)+[a-zA-Z]{2,63}?$')) return(true);
```
stdweird: simpler:
```pan
return(match(...));
```

pan/types.pan
```diff
@@ -837,24 +851,25 @@ function is_network_name = {
...
+    if (is_hostname(ARGV[0])) return(true);
+
+    if (is_shorthostname(ARGV[0])) {
+        deprecate(0, "Short hostnames are deprecated as valid network_names.");
+        return(true);
     };
 
+    if (is_top_level_domain(ARGV[0])) return(true);
+
     # Not a hostname. Is it a IP/mask?
```
stdweird: move this to its own function/type combo?

jrha: What do you mean? I did add `type_top_level_domain`.

stdweird: i meant this block of code `is_ip_and_mask`? (i know it's not your code and all that, so feel free to ignore)

jrha: Oh ok, sure.

pan/types.pan
```diff
@@ -829,6 +829,37 @@ type type_uppercase = string with {
...
+    is_top_level_domain(SELF);
+};
+
+
+function is_ipv4_netmask_pair = {
+    pair = split('\/', 1, ARGV[0]);
+    if (length(pair) == 2) {
+        ip = pair[0];
+        netmask = pair[1];
+        if (is_ip(ip) && (is_ip(netmask) || match(netmask, '^(\d+)$'))) {
```
stdweird: maybe `(match(netmask, '^(\d+)$') && to_long(netmask) >= 0&&  to_long(netmask) <= 32)`

jrha: This bit conflates netmasks and prefix lengths, perhaps it is worth defining types for them?

stdweird: sure, seems like a nice addition.

pan/types.pan
ned21: Drop the "return" and just say: ```match(ARGV[0], ...);```

As per PAN compiler doc, return() should be avoided if possible for performance reasons.

pan/types.pan
```diff
@@ -829,6 +829,37 @@ type type_uppercase = string with {
...
 
 @documentation{
+Checks if the argument is a valid top level domain name (e.g. .example)
+}
+function is_top_level_domain = {
+    return(match(ARGV[0], '^(\.)+[a-zA-Z]{2,63}?$'));
+};
+
+
+type type_top_level_domain = string with {
```
ned21: Why the separate function call? 

```
type type_top_level_domain = string with match(...); 
```

pan/types.pan
```diff
@@ -829,6 +829,37 @@ type type_uppercase = string with {
...
+function is_ipv4_netmask_pair = {
+    pair = split('\/', 1, ARGV[0]);
+    if (length(pair) == 2) {
+        ip = pair[0];
+        netmask = pair[1];
+        if (is_ip(ip) && (is_ip(netmask) || match(netmask, '^(\d+)$'))) {
+            return(true);
+        };
+    };
+    return(false);
```
ned21: s/return(false)/false/ for performance

pan/types.pan
```diff
@@ -837,24 +868,20 @@ function is_network_name = {
...
-    };
-    # Not IP/mask: is it .domain?
-    if (index ('.', hst) == 0) {
-        return (is_hostname (substr (hst, 1)));
-    };
+    is_ipv4_netmask_pair(ARGV[0]);
+
     # Everything failed!
-    return (false);
+    return(false);
```
ned21: Drop return() can just say "false"

pan/types.pan
jrha: The function is for use in `is_network_address`, the type is a freebie.

pan/types.pan
jrha: Neat, didn't know that.

pan/types.pan
stdweird: @jrha if you have no time for this atm, simply open an issue for it so this can be merged.

jrha:
@stdweird: updated

pan/types.pan
```diff
@@ -829,6 +829,47 @@ type type_uppercase = string with {
 
 
 @documentation{
+Checks if the argument is a valid top level domain name (e.g. .example)
+}
+function is_top_level_domain = {
+    return(match(ARGV[0], '^(\.)+[a-zA-Z]{2,63}?$'));
```
stdweird: remove the `return`

stdweird:
@jrha one more tiny remark

stdweird:
@jrha more LGTM than ever

PR #80, opened 2015-09-14
Blockdevices: add metadata option to md devs

kwaegema:
Comes with PR https://github.com/quattor/ncm-lib-blockdevices/pull/56

### DIFF
--- a/quattor/blockdevices.pan
+++ b/quattor/blockdevices.pan
@@ -53,6 +53,7 @@ type blockdevices_md_type = {
     "raid_level" : string with match (SELF,'^RAID[0156]$')
     "stripe_size" : long = 64 # "Stripe size in KB"
     "num_spares" ? long # "Number of spare devices"
+    "metadata" ? string = '0.90' with index(SELF, list('0', '0.90', '1.0', '1.1', '1.2', 'ddf', 'imsm', 'default')) > 0
 };
 
 @documentation{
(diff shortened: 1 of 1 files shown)


stdweird:
LGTM

quattor/blockdevices.pan
```diff
@@ -53,6 +53,7 @@ type blockdevices_md_type = {
     "raid_level" : string with match (SELF,'^RAID[0156]$')
     "stripe_size" : long = 64 # "Stripe size in KB"
     "num_spares" ? long # "Number of spare devices"
+    "metadata" : string = '0.90' with match(SELF, '^(0(.90)?|1(.[012])?|ddf|imsm)$')
```
ned21: The dots needs to be escaped. This appears to be matching the spec from http://linux.die.net/man/8/mdadm which says that the default is 1.2 _not_ 0.90. Perhaps this should be made optional instead of defaulting to an older version?

Using a regexp to fake an ENUM type is pretty ugly. How about using a list instead?
```with index(SELF, list('0', '0.90', '1.0', '1.1', '1.2', 'ddf', 'imsm')) > 0;``` That probably results in a user-unfriendly error message so still not idea.

stdweird:
retest this please

jrha:
What on earth does this do?
Can you add/link to some documentation please?

kwaegema:
@jrha comment added

PR #82, opened 2015-09-17
Fix problems with is_network_name and related functions

jrha:
Somehow these slipped through my prior testing and were only exposed when running `ncm-metaconfig` unit tests. :disappointed:


jrha:
Unit tests will still fail on this repository because there are lots of other problems with it :hankey:

stdweird:
LGTM

PR #83, opened 2015-09-17
Distinguish between IPv4 and IPv6 networks

jrha:
Permit both in `is_network_name`, fixes the test failures seen in quattor/configuration-modules-core#582


stdweird:
LGTM

PR #85, opened 2015-09-25
quattor client ncm-ncd rpm do not overwrite lock

### LINKED ISSUE #84: quattor/client/rpms.pan: fix ncm-ncd
```diff
+    ncd = escape('ncm-ncd');
+    if(! exists(ncd)) {
+        SELF[ncd] = nlist();
```

kwaegema:
fixes #84

### DIFF
--- a/quattor/client/rpms.pan
+++ b/quattor/client/rpms.pan
@@ -21,7 +21,6 @@ include { 'rpms/package_default_versions' };
     # Quattor
     SELF[escape('cdp-listend')] = nlist();
     SELF[escape('ncm-cdispd')] = nlist();
-    SELF[escape('ncm-ncd')] = nlist();
     SELF[escape('ncm-query')] = nlist();
 
     # ideally, include components/spma/config with version locked rpm
@@ -29,6 +28,11 @@ include { 'rpms/package_default_versions' };
     if(! exists(spma)) {
         SELF[spma] = nlist();
     };
+    ncd = escape('ncm-ncd');
+    if(! exists(ncd)) {
+        SELF[ncd] = nlist();
+    };
+
       
     SELF;
 };


quattor/client/rpms.pan
```diff
@@ -29,6 +28,11 @@ include { 'rpms/package_default_versions' };
     if(! exists(spma)) {
         SELF[spma] = nlist();
     };
+    ncd = escape('ncm-ncd');
```
stdweird: why not put all of them in a list and loop over them?

ned21: The logic---"add if not present but do not overwrite a version lock"---is the purpose of pkg_add() so you can just use that instead. Unless the comment on line 26 is suggesting that components/spma/config cannot be included here?

kwaegema: Hmm, in the docstring of pkg_add I read : If the 'multi' option is not set, the pkg_add function will fail if a previous version is found in the profile. (Multi option is for multiple versions running together, like kernel). so that does not look the same.

ned21: That was how pkg_add() used to work, in my initial note I wrote pkg_repl() instead which used to do the right thing. However upon reviewing the most recent function definitions I came to the conclusion that pkg_add() should also now work, but maybe I am wrong. 

Either way, we should have a function that does this rather than hacking it into each piece of code like this. :-)

piojo: `multi` doesn't work anymore, IIRC. In any case, `pkg_add` or `pkg_repl` should be the way to go here, indeed, @kwaegema, can you re-check?

stdweird:
retest this please

quattor/client/rpms.pan
kwaegema: But can pkg_add or pkg_repl be used without defined version? If I try that, I get an error:
``user-initiated error: no default package version defined for package: ncm_2dncd``

jrha:
@stdweird is this ready?

ned21:
@jrha - no. The functionality implemented here is definitely very useful and should be provided by one of the functions defined by spma. It seems we may not currently have such a function so the discussion needs to move to how/where that should be done!

stdweird:
@ned21 
* https://github.com/quattor/template-library-core/blob/master/quattor/functions/package.pan#L82, but with support for missing version (i'm not sure you can pass undef as version to `pkg_update`, but modifying it would make it backwards incompatible)
* `pkg_aonly` as a mix of `pkg_add` and `pkg_ronly`, to add only in case it's not already present.

ned21:
It looks like kg_update doesn't error check for a version arg but then doesn't work properly if one is not supplied. But underneath that just calls pkg_add which @kwaegema reported as requiring  version so pkg_add probably needs fixing anyway...

kwaegema:
I'm willing to fix these functions. but I'm not yet sure what functionality we all want in which functions.

ned21:
Functions I would like to be able to use:
1. Ensure a package is installed, any version.
2. Install a specific version of a package. Return an error if someone else has already specified a different version of the same package.
3. Install package X, version N or higher. Error if already pinned at something lower, but no-op if existing pinned version if ≥N

(3 may be hard to do in PAN given rpm's "interesting" versioning semantics.)

stdweird:
looks like a mistake in the UGent templates. we should use `pkg_repl`

kwaegema:
there is a difference running pkg_repl with version undef and empty string, the former doing the trick. will redo this pr with pkg_repl

jrha:
@kwaegema any update?

kwaegema:
@jrha  Well, i retried, but seeing something else now: If I do pkg_repl or pkg_add with an undef version, and it was already defined, I get an error , resp.  ``user-initiated error: Attempt to unlock version of package ncm-ncd (version 16.2.0-rc3_SNAPSHOT20160217165040)`` and   ``user-initiated error: Package ncm-ncd is already part of the profile (existing version=16.2.0-rc3_SNAPSHOT20160217165040, requested version=undef)``
So afaik, there is not a function that can 'add a package with undefined version, but keep the version if it already is specified elsewhere' . Can someone confirm this? I can add this function ..

kwaegema:
superseded by #114

PR #87, opened 2015-09-29
Add a benchmark field to the hardware type definition

apdibbo:
This is intended for use with batch systems in order to calculate scaling factors for accurate wall time calculations
I could also see this being useful for other systems such as accounting for storage systems.

### DIFF
--- a/quattor/types/hardware.pan
+++ b/quattor/types/hardware.pan
@@ -6,7 +6,7 @@ include 'quattor/types/annotation';
 include 'quattor/physdevices';
 
 @documentation{
-    RAM definition    
+    RAM definition
 }
 type structure_ram = {
     include structure_annotation
@@ -57,18 +57,18 @@ type structure_rack = {
     Fiber channel definition
 }
 type fcahwaddr = string with is_a_fcahwaddr (SELF);
-type structure_fca = { 
-    include structure_annotation 
-    "hwaddr" ? fcahwaddr # "IEEE fiber Channel World Wide Name" 
-    "active" ? boolean # "Is this port being used" 
+type structure_fca = {
+    include structure_annotation
+    "hwaddr" ? fcahwaddr # "IEEE fiber Channel World Wide Name"
+    "active" ? boolean # "Is this port being used"
 };
 
 @documentation{
     BMC controller
 }
-type structure_bmc = { 
-    include structure_annotation 
-    "hwaddr" ? type_hwaddr 
+type structure_bmc = {
+    include structure_annotation
+    "hwaddr" ? type_hwaddr
 };
 
 @documentation{
@@ -177,17 +177,17 @@ type structure_sysloc = {
 
 @documentation{
     System benchmark results
-    benchmarks is used to hold the performance benchmark for the machine 
+    benchmarks is used to hold the performance benchmark for the machine
     i.e. HEPSpec06 score
     this might be used to scale things such as the wall time
-    example of using this might be : 
-    variable CONDOR_WN_SCALING_FACTOR = value('/hardware/benchmarks/hepspec06') 
+    example of using this might be :
+    variable CONDOR_WN_SCALING_FACTOR = value('/hardware/benchmarks/hepspec06')
           / ( 4 * get_num_of_cores()) ;
 }
 type structure_benchmark = {
-    "hepspec06" ? double
-    "HPL" ? double
-    "stream" ? double
+    "hepspec06" ? double 
+    "HPL" ? double @{unit: Gflops}
+    "stream" ? double @{unit: MB/s TRIAD}
 };
 
 @documentation{
@@ -202,7 +202,7 @@ type structure_hardware = {
     "console"      ? structure_console
     "sysloc"       ? structure_sysloc
     "nodename"     ? string
-    "benchmarks" ? structure_benchmark 
+    "benchmarks" ? structure_benchmark
     # Obsolete field, use the appropriate "cards" sub-field instead!!
     "harddisks"    ? structure_raidport{}
 };
--- a/quattor/types/hardware.pan
+++ b/quattor/types/hardware.pan
@@ -185,9 +185,11 @@ type structure_sysloc = {
           / ( 4 * get_num_of_cores()) ;
 }
 type structure_benchmark = {
-    "hepspec06" ? double 
-    "HPL" ? double @{unit: Gflops}
-    "stream" ? double @{unit: MB/s TRIAD}
+    "hepspec06" ? double
+    @{unit: Gflops}
+    "HPL" ? double
+    @{unit: MB/s TRIAD}
+    "stream" ? double
 };
 
 @documentation{
--- a/quattor/types/hardware.pan
+++ b/quattor/types/hardware.pan
@@ -176,6 +176,21 @@ type structure_sysloc = {
 };
 
 @documentation{
+    System benchmark results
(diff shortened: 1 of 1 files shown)


piojo:
Docs? How would you consume this field?

apdibbo:
In our use case we store the HEPSpec06 result here.

We then consume this with the following:
`variable CONDOR_WN_SCALING_FACTOR = value('/hardware/benchmark') / ( 4 * get_num_of_cores()) ;`
this gives us a value per core which we use for scaling the wall time of jobs.

jrha:
Document that in the schema then?

jrha:
There *could* be an argument for adding this to `structure_annotation`, but could get messy.

apdibbo:
Is the next commit what you are looking for or is there another format/location you would prefer?

apdibbo:
please test this

Pansanel:
It would be a nice feature. It could be easily used in:
https://github.com/quattor/template-library-grid/blob/umd-3/personality/wn/machine-features.pan
to define hs06 per slot (it is actually manually defined through `WN_HS06` variable).

apdibbo:
@Pansanel that would be a perfect use for this. Looking at that it also occurs to me that this could be used to populate the MachineFeatures required for the workernodes.

quattor/types/hardware.pan
```diff
@@ -186,6 +186,10 @@ type structure_hardware = {
     "console"      ? structure_console
     "sysloc"       ? structure_sysloc
     "nodename"     ? string
+    #benchmark is used to hold the performance benchmark for the machine i.e. HEPSpec06 score
+    #this might be used to scale things such as the wall time
+    #example of using this might be : variable CONDOR_WN_SCALING_FACTOR = value('/hardware/benchmark') / ( 4 * get_num_of_cores()) ;
```
piojo: Almost there. Please keep lines below 80 characters

quattor/types/hardware.pan
```diff
@@ -186,6 +185,21 @@ type structure_hardware = {
...
+    # this might be used to scale things such as the wall time
+    # example of using this might be : 
+    # variable CONDOR_WN_SCALING_FACTOR = value('/hardware/benchmark/HEPSpec06') 
+=======
+    #benchmarks is used to hold the performance benchmark for the machine 
+    #i.e. HEPSpec06 score
+    #this might be used to scale things such as the wall time
+    #example of using this might be : 
+    #variable CONDOR_WN_SCALING_FACTOR = value('/hardware/benchmark/HEPSpec06') 
+>>>>>>> origin/master
```
piojo: Conflict leftovers... Do **not** merge master into your topic branches.  The workflow is topic -> master, never the other way around.

stdweird:
test this please

quattor/types/hardware.pan
```diff
@@ -186,6 +185,21 @@ type structure_hardware = {
...
+    #       / ( 4 * get_num_of_cores()) ;
+    "benchmarks"    ? double{}
```
stdweird: can you make a benchmark type with the name of each benchmark? (also add `HPL` and `stream`)

jrha: Do we really want to control the benchmark names?
Would it be enough to enforce a naming pattern?

stdweird: you like typos much? i know that adding site specific tests and tus naming can clutter the schema, but this is not ok imho.
we can ship a pretty empty `quattor/types/benchmarks.pan` and let the site decide/override?

jrha: You know I don't like typos, which is why I keep correcting yours :stuck_out_tongue:
But everything we bring into the schema increases our maintenance overhead, so I have to ask.

stdweird: :smile: 
i'm aware, but i don't see what is wrong with a list of known benchmark names (maybe with url/smalldescription), also wrt units used in said benchmarks.

jrha:
test this please

jrha:
test this please

jrha:
@stdweird any comments?

quattor/types/hardware.pan
```diff
@@ -175,6 +175,21 @@ type structure_sysloc = {
 };
 
 @documentation{
+    System benchmark results
+    benchmarks is used to hold the performance benchmark for the machine 
+    i.e. HEPSpec06 score
```
stdweird: `i.e.` -> `e.g.`

quattor/types/hardware.pan
```diff
@@ -175,6 +175,21 @@ type structure_sysloc = {
...
+    this might be used to scale things such as the wall time
```
stdweird: i'm not sure the example adds much (the variable is very specific, and this is core templates after all). but i can't come up with a better example, so feel free to leave it.

quattor/types/hardware.pan
```diff
@@ -175,6 +175,21 @@ type structure_sysloc = {
...
+    example of using this might be : 
+    variable CONDOR_WN_SCALING_FACTOR = value('/hardware/benchmarks/hepspec06') 
+          / ( 4 * get_num_of_cores()) ;
+}
+type structure_benchmark = {
+    "hepspec06" ? double
```
stdweird: maybe add a `@{unit}`, do describe what is being measured/reported.

quattor/types/hardware.pan
```diff
@@ -175,6 +175,21 @@ type structure_sysloc = {
...
+    "HPL" ? double
```
stdweird: add `@{unit: Gflops}`

quattor/types/hardware.pan
```diff
@@ -175,6 +175,21 @@ type structure_sysloc = {
...
+    "stream" ? double
```
stdweird: `@{unit: MB/s TRIAD}`

stdweird:
@apdibbo i think a adding a small docstring indicating the unit used might help

quattor/types/hardware.pan
jrha: Is the stream definition too simplistic? Should it be named `stream-triad` or similar?

stdweird: stream is 4 benchmarks, so if you want all of them, i guess a a substructure is even better, but on the other hand it gets too complicated. i think settling on one value is enough.

jrha:
@apdibbo this is almost ready to be merged, can you rebase and add the unit docstrings as suggested by @stdweird?

apdibbo:
test this please

stdweird:
@apdibbo @jrha need to check what happens with the annotation when they are added on same line but after the attribute (in case they are added to the next one)

jrha:
@apdibbo unfortunately @stdweird is correct, I've just tested this and the units get attached to the *next* property, as this stands this means, for example `stream` gets `unit: Gflops`.

Could you move the annotations to the line above each propery? (sorry)

jrha:
Cool, that works correctly now.

jrha:
@stdweird do we have a standard for fields appearing in the annotations? It occurs to me that having the units as a specific field might be useful.
i.e.
* `@{unit: Gflops}` gives `<desc>unit: Gflops</desc>` (what we have now)
* `@{unit = Gflops}` gives `<unit>unit: Gflops</unit>`

stdweird:
hmm, i thought you needed `@unit{}` for that, nice to know `@{unit = }` also works.
wrt standards, there's only https://xkcd.com/927/

jrha:
test this please

PR #90, opened 2015-10-07
new function template

sartiran:
Hi,

as it was needed for some grid storage configuration, I wrote a function comparing versions numbers (a simple string comparison does not do the jobs). I guess this may be a function of general interest so it would be nice to have in under some standard path. I'm not sure this (quattor/functions) is the correct location where to put the template, so, please let me know which is the best way to proceed.

Cheers,
a.


stdweird:
test this please

jrha:
test this please

quattor/functions/strings.pan
```diff
@@ -0,0 +1,29 @@
+declaration template quattor/functions/strings;
+
+include { 'pan/functions' };
```
jrha: Remove the `{ }`

jrha:
This looks good, it could do with some documentation to explain how it "compares" versions.
Add documentation comments as [pan annotations](https://quattor-pan.readthedocs.org/en/pan-10.2/pan-book/pan-book.html#annotations).

quattor/functions/strings.pan
```diff
@@ -0,0 +1,29 @@
...
+  for(i=1;i<5;i=i+1){
+    if(to_long(version1[i])>to_long(version2[i])){
+      return(-1);
+    };
+    if(to_long(version1[i])<to_long(version2[i])){
+      return(1);
+    };
+  };
+
+  return(0);
```
jrha: You can leave out the `return()` here and just assert `0`

jouvin: in fact you must leave out the `return()`... Having an explicit `return()` is handled internally as an exception adn breaks the parallelism into the compiler... `return` must be used only when it is required to break the normal execution flow... (in fact there are quite a lot of useless `return` in the template library coming from the ancient times...)

quattor/functions/strings.pan
```diff
@@ -0,0 +1,29 @@
+declaration template quattor/functions/strings;
+
+include { 'pan/functions' };
+
+#version comparison function (should not be here)
```
stdweird: add this as documentation using 
```pan
@documentation{
    function, arguments,...
}
```

quattor/functions/strings.pan
```diff
@@ -0,0 +1,29 @@
...
+function comparever = {
+  if(ARGC!=2){
```
stdweird: spaces around `=`

quattor/functions/strings.pan
```diff
@@ -0,0 +1,29 @@
...
+    error("comparever: wrong number of arguments. Required 2, given "+to_string(ARGC)+".");
```
stdweird: use format, like 
```pan
error(format("comparever: wrong number of arguments. Required 2, given %s.", to_string(ARGV));
```
(this also prints the actual arguments, not only the number , to make clear what is passed)

quattor/functions/strings.pan
```diff
@@ -0,0 +1,29 @@
...
+  };
+  version1=matches(ARGV[0],'(\d+)\.+(\d+)\.+(\d+)[-]*(\d*)');
```
stdweird: the pattern should be `^(\d+)\.(\d+)\.(\d+)(?:-(\d+))? ` or even `^(\d+)\.(\d+)(\.(\d+)(?:-(\d+))?)?`

and is best that you store the pattern in a variable so you are sure it's same pattern being matched.

but the remainder of the logic than also changes (a lot).

so maybe (untested)
```pan
pattern = '^(\d+)\.(\d+)(\.(\d+)(?:-(\d+))?)?';
version1 = matches(ARGV[0], pattern);
version2 = matches(ARGV[1], pattern);
len = length(version1);
if(len != length(version2)) {
    error(format("Compared versions are of different format (supported %s): %s vs %s", pattern, ARGV[0], ARGV[1]));
}
if (len == 0) {
    error(format("Compared versions do not match format %s: %s , %s", pattern, ARGV[0], ARGV[1]));
}
for (i=1; i <= len; i=i+1) {
...
} 
```

quattor/functions/strings.pan
```diff
@@ -0,0 +1,29 @@
+declaration template quattor/functions/strings;
+
+include { 'pan/functions' };
+
+#version comparison function (should not be here)
+function comparever = {
```
stdweird: @sartiran i just found https://github.com/quattor/template-library-core/blob/master/quattor/functions/package.pan#L4, is that something you could reuse?

sartiran: Yes of course I can! That was exactly what I was looking for before deciding to write one of my own. I've tested it and I can reuse the pkg_compare_version function as it is for the moment (and, if later needed, I'll modify it).  I think we can just cancel this pull request and I'll modify the templates in the the pull request #154 so that they use the pkg_compare_version function

jrha:
No longer required.

PR #91, opened 2015-10-08
Make system kernel version optional

jrha:
I'm not sure why it was required, we certainly have no problems on systems where it is not set.

### DIFF
--- a/quattor/types/system.pan
+++ b/quattor/types/system.pan
@@ -46,7 +46,7 @@ type structure_system = {
     "filesystems"   ? structure_filesystem[]
     "blockdevices" ? structure_blockdevices
     "glite"         ? structure_glite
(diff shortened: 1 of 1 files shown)


ned21:
See https://github.com/quattor/configuration-modules-core/issues/237 and https://github.com/quattor/configuration-modules-core/pull/231

jrha:
Thanks. Hmm, we're not using `ncm-grub`.

jrha:
Fine, I guess we can just use the "magic" `YUM-managed` value, but it's not ideal.

ned21:
I recently came across a case where /system/kernel appeared to be required by AII so I think making /system/kernel required and the sub-attributes optional is the way to go. Any reason this was closed?

PR #102, opened 2016-02-04
Add `is_choice` to test entry in list

stdweird:
Use case is in schema files to use

``` pan
    "attribute" ? string with is_choice("a", "b", "c")
```

instead of current `with match(SELF, ''^(a|b|c)$")`

(and it also allows `string[] with is_choice(...)` transparently)


stdweird:
This PR is opened without any actual `is_choice` code
@jrha the .`tests` location is temporary

pan/functions.pan
```diff
@@ -44,85 +26,80 @@ function push = {
         error("push can only be applied to a list");
     };
 
-    # Merge the arguments into the given array.  Neither the
-    # first/next or merge functions can be used because the
-    # ARGV array cannot be directly referenced.
+    # Cannot use merge or refernce ARGV directly
     i = 0;
     while (i<ARGC) {
         v[length(v)] = ARGV[i];
```
ned21: I think this can be replaced with append(ARGV[i]) ? That's what I deployed to our code base internally recently.

pan/functions.pan
```diff
@@ -44,85 +26,80 @@ function push = {
...
+    "/data" = push_if(CONDITION, 2, 3);
+
+    "/data" will contain list(1, 2, 3)
+}
 function push_if = {
     # Check cardinality and first argument.
-    if (ARGC == 0 || !is_boolean(ARGV[0]))
-        error("usage: \"/full/path\" = push(boolean, values...)");
+    if (ARGC == 0 || !is_boolean(ARGV[0])) {
+        error("usage: push(boolean, values...) requires at least a boolean as first argument");
```
ned21: s/push/push_if/

pan/functions.pan
```diff
@@ -156,115 +133,126 @@ function push_if = {
...
+function host_domain_from_object = {
     # Check cardinality.
-    if (ARGC != 0) error("usage: hostname_from_object()");
+    if (ARGC != 1) {
+        error("usage: host_domain_from_object(default_domain) requires default domain as argument");
+    };
+
+    # Check the argument to ensure it is a valid domain name.
+    if (!is_fqdn(ARGV[0])) {
+        error(format("host_domain_from_object: invalid default domain (%s)", ARGv[0]));
```
ned21: ARGv[0] => ARGV[0] ?

stdweird:
@ned21 from the comments in the old code, it looked that `ARGV` was soem special construct and couldn't be used with `merge` etc. however, that doesn't seem to be the case in pan 10.2 (and also stated as such in the pan docs). see https://github.com/quattor/template-library-core/commit/b1b170c2dd489124b6d3de13b5ffc00e46075f82

what release of panc can we use to releas the core templates?
also, can you have a good look at the unittests to see what can be added to make sure the new code is correctly behaving?

pan/functions.pan
```diff
@@ -1,270 +1,290 @@
...
+    Example:
+    object template myhost;
+    "/data" = hostname_from_object("example.net");
+
+    "/data" will contain "myhost.example.net"
+}
 function full_hostname_from_object = {
     # Check cardinality; leave detailed error checking to called functions.
-    if (ARGC != 1) error("usage: full_hostname_from_object(default_domain)");
+    if (ARGC != 1) error("usage: full_hostname_from_object(default_domain) requires default_domian as argument");
```
jouvin: default domian -> default domain

pan/functions.pan
```diff
@@ -1,270 +1,290 @@
...
-    if (ARGC == 0 || !is_boolean(ARGV[0]))
-        error("usage: \"/full/path\" = push(boolean, values...)");
+    if (ARGC == 0 || !is_boolean(ARGV[0])) {
+        error("usage: push_if(boolean, values...) requires at least a boolean as first argument");
+    };
 
     if (ARGV[0]) {
-        # Get the reference to SELF or create an empty list
-        # as necessary.
+        delete(ARGV[0]);
```
jouvin: Why deleting `ARGV[0] `?

pan/functions.pan
```diff
@@ -1,270 +1,290 @@
...
 };
 
+@documentation{
+    Test if (all elements of) the first argument are
+    any of the remaining arguments.
+    (If the first argument is a list, all its elements are tested).
+
+    Example:
+    is_choice_of(1, 1, 2, 3) returns true
+    is_choice_of(1, 2, 3) returns false (a string is not a long)
```
jouvin: Not sure to understand the comment `a string is not a long`...

pan/functions.pan
```diff
@@ -1,270 +1,290 @@
...
+
+    is_choice_of(list("b", "c"), "a", "b", "c") returns true
+    is_choice_of(list("a", "d"), "a", "b", "c") returns false ("d" is not an argument)
```
jouvin: `"d" is not an argument` -> `"d" is not one of the valid values`

pan/functions.pan
```diff
@@ -1,270 +1,290 @@
...
+
+    The types are not checked and are assumed to be comparable.
+        is_choice_of("1", 1, 2, 3) will give a panc error (string vs long),
```
jouvin: Not very good... For me this is something to avoid. The function should return an understandable message in case of type mismatch. Personnally I'll enforce type checking, this wil add a real value to `is_choice_of()`.
* First check that all valid values passed have the same type
* Second check that any value to check if of this type

jouvin:
@stdweird about `merge()` and `ARGV`, my guess it that this is a very old comment (this template if probably coming from ancient times without being significantly cleaned up...). As for the minimal panc version required by the template library (not only the core part), we had a discussion about this in the last year. I was not able to find the issue where it was discussed but I think we agreed than 10.2 was the minimum version required (as there is no reason to stay with 10.0 or 10.1 and 10.2 was brinig important improvements for metaconfig (to help support of variable names).

pan/functions.pan
stdweird: because it is the boolean condition, not a value to push (old code started looping from 1)

pan/functions.pan
stdweird: nothing to understand, just plain wrong

pan/functions.pan
stdweird: i know, but how do you compare or even get the type of a variable in pan?

jouvin: I don't know other solution that "brute force" (may be worth an issue in panc to have a function returning the type). In this particular case, my solution would be:
- Test all type possibilities with the second argument and save its type
- For all other arguments, test if they are the same time, else raise an error. for this test, you probably need another `if` with all possibilities (a function?).

pan/functions.pan
```diff
@@ -1,270 +1,290 @@
...
-## @param:value value for the entry
-## @example
-##+# "/data" will contain nlist("one",1,"two",2,"three",3)
-##+"/data" = nlist("one",1);
-##+"/data" = npush("two",2,"three",3);
-##=
-############################################################
+@documentation{
+    pushes zero of more pairs (key, value) into a
+    dict.  If the list does not exist or is not defined
```
ned21: s/list/dict/ in this line

unittests/build_all_templates.sh
```diff
@@ -0,0 +1,142 @@
...
+
+# for quattor/client/config
+variable QUATTOR_PROFILE_URL = "https://server.local.domain/profiles";
+variable FULL_HOSTNAME = 'test.local.domain';
+
+# template metaconfig/ganesha/config_v1 has it set to undef (which is ok)
+variable GANESHA_FSAL = 'gpfs';
+
+# can't be undef
+"/software/repositories" = append(dict(
```
ned21: ```"/software/repositories" ?= list(dict());``` would be better?

unittests/errors/pan/functions/host_domain_from_object3.pan
```diff
@@ -0,0 +1,6 @@
+object template host_domain_from_object3;
+
+include 'pan/functions';
+
+@{fails with invalid shorthostname due to _ in name}
```
ned21: This is excellent stuff! We should definitely be able to start using this framework. One more test I can think of, for npush - making sure that a duplicate key still fails. 

```
"/a" = dict('a', true);
"/a" = npush(''a", false);
```

stdweird: the error is thrown by merge in that case, so i will also allow `evaluation` errors

unittests/build_all_templates.sh
stdweird: maybe, but doesn't matter that much in this test.

jouvin:
Any good reason for this PR not to be merged yet... Would be good to have in 16.10 so that we can start using it,..

stdweird:
this PR is completely irrelevant if someone could be bothered to review merge https://github.com/quattor/pan/pull/121 and https://github.com/quattor/pan/pull/125 ...

jrha:
Those two PRs are now merged, so this probably needs reworking.

jrha:
@stdweird is now irrelevant?

jrha:
@stdweird ping?

PR #108, opened 2016-03-09
adding HW tags

sartiran:
Very small change proposed. This is just to request the possibility to have user defined HW tags by adding a "tags" nlist to annotation.pan. We've found useful at LLR to have the possibility to have some suitably defined tags associated to some hardware profile and use them to trigger some configurations/features.

### DIFF
--- a/quattor/types/annotation.pan
+++ b/quattor/types/annotation.pan
@@ -19,4 +19,5 @@ type structure_annotation = {
    "lang"         ? string # "language of the product"
    "power"        ? long   # "power in watts"
    "location"     ? string # "location of the hardware"
+   "tags"         ? nlist  # "user defined HW tags"
 };
--- a/quattor/types/annotation.pan
+++ b/quattor/types/annotation.pan
@@ -19,5 +19,5 @@ type structure_annotation = {
    "lang"         ? string # "language of the product"
    "power"        ? long   # "power in watts"
    "location"     ? string # "location of the hardware"
-   "tags"         ? nlist  # "user defined HW tags"
+   "tags"         ? dict   # "user defined HW tags"
 };


quattor/types/annotation.pan
```diff
@@ -19,4 +19,5 @@ type structure_annotation = {
    "lang"         ? string # "language of the product"
    "power"        ? long   # "power in watts"
    "location"     ? string # "location of the hardware"
+   "tags"         ? nlist  # "user defined HW tags"
```
stdweird: use `dict`

jrha:
please test this

jrha:
Can you give some examples of the potential usage of this field?

sartiran:
For example:
"/hardware/tags" = nlist(
  'ipmi', 'XX' # <--- different HW types have different control procs and this has implications on the
                  #        we configure ipmi, etc
  'type', 'XX' # <-- there are HW type distinctions that go beyond the granularity of vendor/model
  'raid', 'XX'  # <-- different types of ride cards and config require, for example, different probes for 
                   #       testing, etc..
);

well, everything can be probably done by hijacking a bit the annotation fields that are already there. But I  think is cleaner (and more flexible) to have some user definable tags.

jrha:
I'm not sure this is a good solution for those cases.

There is already a type field, which we use for greater granularity at RAL.
For ipmi and raid, you would be better off using structures under `cards`, e.g. (copied from real templates — no warranty!)...

#### In the structure template for each model:
```pan
'type' = 'storage';

'cards/raid/_1'= dict(
    'manufacturer', 'adaptec',
    'model', '52445',
    'bus', 'PCI-E',
    'location', 'slot 4',
    'bbu', true,
    'numberports', 24,
    'cache', 512*MB,
    'ports', dict(),
);

'cards/bmc' = list(
    dict(
        'manufacturer', 'supermicro',
    ),
);
```

#### Somewhere common to most hosts:
```pan
include {
    if (exists("/hardware/cards/bmc")) {
        'site/config/ipmi_install'
    };
};

# Install IPMI ganglia metrics
include {
    if (exists("/hardware/cards/bmc")) {
        'monitoring/ganglia/sensors/gmetric-ipmi'
    };
};

# Do something specific for nodes with dell BMCs
include {
    if ( exists('/hardware/cards/bmc/0/manufacturer') ) {
        if (to_lowercase(value ('/hardware/cards/bmc/0/manufacturer')) == 'dell' ) {
            'dark_magic'
        };
};

variable HW_RAID_CARD ?= "none";
variable HW_RAID_CARD = {
  if (path_exists("/hardware/cards/raid/_1/manufacturer") && is_defined(value("/hardware/cards/raid/_1/manufacturer"))) {
    return(to_lowercase(replace(" ", "_", value("/hardware/cards/raid/_1/manufacturer"))));
  }
  else {
    return(SELF);
  }
};
```

sartiran:
These are A LOT of lines for adding a tag.... and it only moves the problem because, still, you are dealing with pre-defined attributes. 

The type field is a string, if you have more than 1 information to encode in it, you can, for example, use it as a comma separated list of tags and parse it. But having a real  tag list is way much cleaner. 

On a general ground, and IMHO, if you have a list of annotation, having a "custom" field is a good idea. You cannot know "a priori" everything that someone needs to annotate. What if I want to annotate the color of my machines ... 
On a more realistic ground what if the way I operate a piece of HW depends on the expiration date of its guarantee, the budget line it was used to purchase it, a known HW failure that I cannot fix but I can work around .... I should be able to attach to the HW definition everything I see as connected to the HW itself. 

But ok  this 1 line of code has already required me too many lines of comments :) . We can keep on using it by overriding the template, so if you think this is not in the line with the general quattor policies (which makes perfectly sense), just reject the PR.

jrha:
I don't disagree in principal, but for the example you gave I think there are better existing solutions.

Sorry if my examples weren't clear, but only the first block of code adds tags, I also included examples of how we make use of the tags to drive configuration.

A very minimal example to add tags to a model template might be something like the following:
```pan
'cards/raid/_1/manufacturer' = 'adaptec';
'cards/bmc/0/manufacturer' = 'supermicro';
```
Which would be enough to (for example) install the correct vendor utilities for each device.

jrha:
We discussed this at the 21st workshop, our feeling was that you should be able to accomplish this with what already exists in the schema, but we may not have understood your use cases. Could you give us some real world examples, we are happy to extend the schema if needed but we're not convinced yet.

sartiran:
Hi Adam, 

indeed it doesn't matter. I managed to do what I had to do adding the different HW components in the proper way, as you suggested. I'm still convinced that having a user defined set of tags could give more flexibility to the tool. The point is that I'd like to have a quick and dirty way to cope with unforeseen cases. As a real world example, imagine that you want to tag machines (or HW devices inside machines) which are old and no longer under guarantee. For example you may want to do so in order to trigger the configuration of a more detailed status monitoring on these machines.

PR #111, opened 2016-04-11
Add new template for legacy types

jrha:
Types in here should be collected to be phased out of use.
The first is the infamous non-boolean yesnostring used by several component with different names.


stdweird:
thx!

PR #94, opened 2015-11-13
Add boolean hyperthreading for structure cpu.

rhumroux:
Hi,

 I'm a new quattor user from T2_BE_IIHE, this is my first pull request, so I apologize beforehand if I'm doing something wrong !

I'd like to make changes to how slots are calculated for hyperthreaded CPUs. 
This merge request just adds a new field to the cpu structure. It will allow us to compute the number of logical cores for CPUs with hyperthreading. It defaults to false.

Cheers
Romain

PR #113, opened 2016-04-21
Allow thread count of CPU to be specified

jrha:
Following on from #94 and related to quattor/template-library-standard#66, it would be nice to be able to record the physical threadcount of CPUs.

### DIFF
--- a/quattor/types/hardware.pan
+++ b/quattor/types/hardware.pan
@@ -26,7 +26,7 @@ type structure_cpu = {
     "cores" : long(1..) = 1
     @{ desc = Number of execution threads on each CPU chip, e.g. a hyperthreaded eight core chip would have 16 threads }
     "max_threads" ? long(1..)
-    "hyperthreading" : boolean = false
+    "hyperthreading" : boolean = false with {deprecated(0, 'The hyperthreading cpu property has been deprecated, please migrate to max_threads'); true;}
 };
 
 
--- a/quattor/types/hardware.pan
+++ b/quattor/types/hardware.pan
@@ -23,6 +23,7 @@ type structure_cpu = {
     "speed"  : long # "CPU clock speed in MHz"
     # Number of cores on each CPU chip, defaults to 1.
     "cores" : long(1..) = 1
+    "max_threads" ? long(1..)
     "hyperthreading" : boolean = false
 };
 
--- a/quattor/types/hardware.pan
+++ b/quattor/types/hardware.pan
@@ -20,9 +20,11 @@ type structure_ram = {
 }
 type structure_cpu = {
     include structure_annotation
-    "speed"  : long # "CPU clock speed in MHz"
-    # Number of cores on each CPU chip, defaults to 1.
+    @{ desc = CPU clock speed in MHz }
+    "speed" : long
+    @{ desc = Number of cores on each CPU chip }
     "cores" : long(1..) = 1
+    @{ desc = Number of execution threads on each CPU chip, e.g. a hyperthreaded eight core chip would have 16 threads }
     "max_threads" ? long(1..)
     "hyperthreading" : boolean = false
 };


quattor/types/hardware.pan
```diff
@@ -20,9 +20,12 @@ type structure_ram = {
 }
 type structure_cpu = {
     include structure_annotation
-    "speed"  : long # "CPU clock speed in MHz"
-    # Number of cores on each CPU chip, defaults to 1.
+    @{ desc = CPU clock speed in MHz }
+    "speed"  : long
+    @{ desc = Number of cores on each CPU chip, defaults to 1. }
```
ned21: I think "defaults to 1", fails the "obvious-ness" test of a good comment and should be removed.

quattor/types/hardware.pan
```diff
@@ -20,9 +20,12 @@ type structure_ram = {
 }
 type structure_cpu = {
     include structure_annotation
-    "speed"  : long # "CPU clock speed in MHz"
-    # Number of cores on each CPU chip, defaults to 1.
+    @{ desc = CPU clock speed in MHz }
+    "speed"  : long
+    @{ desc = Number of cores on each CPU chip }
     "cores" : long(1..) = 1
+    @{ desc = Number of execution threads on each CPU chip }
```
stdweird: you mean `each CPU core`.
what is the relation with the hyperthreading option? i'm assuming this option is the number of SMT threads
also, what's up with the `desc =`? all annotations like this are descriptions (for functions, it makes sense to distinguish between args and description

jrha: No I don't, I mean per physical CHIP

stdweird: what's the usecase? (and do you mean chip or numadomain)

jrha: Ability to generate configuration for batch systems/virtualisation platforms based on threads.
A hyperthreading boolean is not enough (core:thread ratio is not always 1:2) and as you mentioned in #94, some platforms even allow the ratio to be defined.

My meaning was per physical package (because it's easy data to get hold of) but threads per core would be straightforward enough as well.

stdweird: hmmm, i don't to see how this is a property of a cpu. but i might be missing something. can you give an example of such config file (and in particular why this multiplier is coupled to the cpu and why the multiplier is not application specific).

jrha:
Ok sure, lets look at some specific examples.

Consider the Intel Xeon [E5-2650L](http://ark.intel.com/products/91752/Intel-Xeon-Processor-E5-2650L-v4-35M-Cache-1_70-GHz) which has 14 cores. Each core has two execution units, which allows it to execute 28 parallel threads in parallel.

A [SPARC T5](http://www.oracle.com/technetwork/server-storage/sun-sparc-enterprise/documentation/o13-024-sparc-t5-architecture-1920540.pdf) CPU on the other hand has 8 execution units per core, giving a 16 core chip the ability to execute 128 parallel threads.

These are consequences of the hardware design, no a property of the software running on a given CPU.

Wrote this on a :bus:. isn't the future awesome.

jrha:
In answer to your earlier question, this would therefore be the number of SMT threads.

stdweird:
so we need a max setting, an actual setting and some relation with hyperthreading.
hyperthreading means max 2, and actual threads also 2 (per core). not sure what hyperthreading menas on other non-intel cpus, but we could deprecate it (hypterthreading is true if number of actual threads is more then one)

quattor/types/hardware.pan
jouvin: @jrha I tend to agree with @stdweird. I understand that this can be a useful information but for me this is a site configuration choice, not something related to the HW characteristics. I think we should restrict this part of the schema to what is related to the HW description. What about a site-specific variable to track this kind of information? (this is in fact how it works in the templates of the template library that configures Torque/MAUI).

jrha: The intention was that this records the maximum supported number of threads, which is absolutely hardware specific.

jouvin:
@jhra thanks for the clarification. Sounds raisonable for me. What about `maxThreads` rather than `threads` to make to property usage a little bit clearer. And also improving the desc...

jrha:
Added link to kernel source's explanation of threads, it's better than mine.

PR #117, opened 2016-07-11
[pan/units] added more binary constants

victor-mendoza:
As discussed in quattor/template-library-grid#180.


jrha:
please test this

PR #119, opened 2016-07-19
Add a transitional type to allow components to migrate to real booleans

jrha:
For the 16.8 cycle I plan to migrate as many properties of as many components as possible from using `legacy_binary_affirmation_string` to `transitional_yes_no_true_false`, then migrate them to `boolean` in a following release cycle.

### DIFF
--- a/pan/legacy.pan
+++ b/pan/legacy.pan
@@ -1,3 +1,15 @@
 declaration template pan/legacy;
 
 type legacy_binary_affirmation_string = string with match(SELF, "^(yes|no)$");
+
+function is_yes_no_true_false = {
+    v = ARGV[0];
+    if (is_boolean(v)) return(true);
+    if (is_string(v) && match(v, "^(yes|no)$")) {
+        deprecated(0, 'Legacy yes/no value in use, please migrate to true/false.');
+        return(true);
+    };
+    false;
+};
+
+type transitional_yes_no_true_false = property with is_yes_no_true_false(SELF);
--- a/pan/legacy.pan
+++ b/pan/legacy.pan
@@ -1,7 +1,20 @@
+@documentation{
+Legacy and transitional type definitions
+
+Should not be used for any new components, they are collected here to aid transition to newer types.
+}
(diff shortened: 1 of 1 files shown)


ned21:
I think it would be useful to have the plan described here in the PR codified in an annotation for the type too so people can understand the intent without needing to open up GH?

jrha:
Good idea!

pan/legacy.pan
```diff
@@ -1,3 +1,12 @@
 declaration template pan/legacy;
 
 type legacy_binary_affirmation_string = string with match(SELF, "^(yes|no)$");
+
+function is_yes_no_true_false = {
+    v = ARGV[0];
+    if (is_boolean(v)) return(true);
+    if (is_string(v) && match(v, "^(yes|no)$")) return(true);
```
stdweird: add some some `deprecated` message?

jouvin: I'd suggest not to add a deprecated message until we are ready to do the effort to do the migration (and first until we evaluated how much work is involved). This would pollute a lot the compilation messages.

stdweird: if we are not going to do anything, why do we even switch to `transitional_yes_no_true_false`?

jouvin: As for me, the first goal is to discourage people to use this inappropriate type... As for migrating the existing use, I think we need a minimum of evaluation about the impact (basically how many components will be affected by a non backward compatible schema)?

stdweird: as soon as we switch to this new type, we have to make changes to the components. but this is backwards compatible in pan, hence i think deprecation warnings are ok.

jrha: My intention here is to actively manage the migration:
1. No components will use this type until their Perl code has been made compatible with real booleans.
2. No component will use this type for more than a single release.

jrha:
My intention is to actively manage the migration:
1. No components will use this type until their Perl code has been made compatible with real booleans.
2. No component will use this type for more than a single release.

PR #121, opened 2016-08-17
Introduce an absolute_file_path type

jrha:
Factors out inner logic valid_absolute_file_paths into new is_absolute_file_path function.

Resolves #116.

PR #120, opened 2016-08-17
Throw error if short hostname is used as a network_name

jrha:
They are not valid, but we have allowed them with a deprecation warning since the 15.8 release.

Closes #78.

PR #110, opened 2016-04-06
aii schema: add protected option

kwaegema:
added value for https://github.com/quattor/aii/pull/172

PR #122, opened 2016-08-17
aii-opennebula: Fix default values of global variables

jrha:
Replace undef with null to allow the schema to apply defaults as the comment suggests

This fixes the test errors seen in #110, #120 and #121.

Also did some re-factoring, because the code made me feel sad.


stdweird:
@alvarosimon we need this cleaned up in aii too

alvarosimon:
@stdweird  ok I will do that asap

Issue #59, opened 2014-12-27
provide `structure_monitoring`


stdweird:
the current `stucture_system` has an optional `monitoring` attribute with type `structure_monitoring`. This type however is not provided as part of this repo, and most likely comes from from the `include 'quattor/monitoring/lemon';`, which is also not part of this repository.

jouvin:
@loomis Any comment on this?

samary:
Any update about this issue ? As we are not using the monitoring repository, we have to comment this line at each template update. Does someone have a proposal to fix this ?

ned21:
For reference this is this [line](https://github.com/quattor/template-library-core/blob/master/quattor/types/system.pan#L30) and this [line](https://github.com/quattor/template-library-core/blob/master/quattor/types/system.pan#L53). If lemon is dead then we can probably just fix do this:

```include { if_exists('monitoring/lemon/schema') };```

but that will still cause the second to fail so we just replace it with a generic nlist.

```type structure_system = {
   ...
   "monitoring"    ? nlist
};
```

If you still use /system/monitoring and want to validate it then you an still do this:
```bind /system/monitoring = structure_monitoring;```

but there is no way with a bind to specify an optional type, bind will convert "? nlist" to ": structure_monitoring". :(

Is anyone still using lemon for whom this would actually be a breaking change?

jouvin:
@ned21 thanks for this proposal, we have been looking how to do it for years! I think `/system/monitoring` was used with Nagios too but I may be wrong...

Commenting on the implementation in [PR](#125).

jrha:
"Fixed" in #125 by removing the type and allowing monitoring templates to bind their own type to the schema.

PR #125, opened 2016-10-06
Make monitoring optional

samary:
Following the proposition of @ned21 (https://github.com/quattor/template-library-core/issues/59#issuecomment-252009328)

### DIFF
--- a/quattor/types/system.pan
+++ b/quattor/types/system.pan
@@ -27,7 +27,7 @@ type structure_oldname = {
 include 'components/network/core-schema';
 
 # TODO: monitoring to be defined
-include 'monitoring/lemon/schema';
+include { if_exists('monitoring/lemon/schema') };
 
 include 'quattor/blockdevices';
 include 'quattor/filesystems';
@@ -50,7 +50,7 @@ type structure_system = {
     "lcg"           ? structure_lcg
     "network"       : structure_network
     # TODO: monitoring-related structures should go elsewhere. 
-    "monitoring"    ? structure_monitoring
+    "monitoring"    ? nlist
     "oldnames"      ? structure_oldname[]
     "rootmail"      : type_email
     "siterelease"   ? string


ned21:
OK to test

stdweird:
LGTM

samary:
Before merging, should I add the type binding in the lemon schema (to be fully transparent and retrocompatible) ?

quattor/types/system.pan
```diff
@@ -27,7 +27,7 @@ type structure_oldname = {
 include 'components/network/core-schema';
 
 # TODO: monitoring to be defined
-include 'monitoring/lemon/schema';
+include { if_exists('monitoring/lemon/schema') };
```
jouvin: I don't completely like the idea that `/system/monitoring` will be bound to a schema based on the existence of a template (I think this lemon template should do the `bind`) as it makes the configuration path no longer optional in the configuration. Instead I'd used a variable that must be defined (I don't remember if we can define global variables in declaration templates) and whose value must be an existing template, like `SYSTEM_MONITORING_SCHEMA`. This would allow to disable this part of the configuration on some machines even if the template exists (by not defining the variable).

jouvin:
I started to play with reviews... Not sure this is a great idea!

jouvin:
@samary yes I think the lemon template should include the `bind` command.. but look at my previous comment.

quattor/types/system.pan
samary: I agree with you. Where do you think this variable should belong ?
I'll update my pull request following your suggestion.

jrha: Does it even need a variable?
Can we just leave it up to sites to include an appropriate schema for `/system/monitoring` if they want one?

jouvin: My point was that the template may exist at a site but you may want not to include it in the context of some machines... For this you need a variable. Thus my proposal to have this variable allowing to specify a template name (namespace) and not to include anything if the variable is undef/null. If we go this way, I'd say that if the variable is defined, the template must exist: this makes troubleshooting easier than adding `if_exists()` but I am open make its existence optional...

I don't remember if a variable can be defined in a declaration template. If yes, we could do:
```
variable SYSTEM_MONITORING_SCHEMA ?= null;
include SYSTEM_MONITORING_SCHEMA;
```

If not, we should do:
```
include if ( is_defined(SYSTEM_MONITORING_SCHEMA ) ) SYSTEM_MONITORING_SCHEMA ;
```

jrha: I don't think that `if_exists` is a good idea either, I just think that sites who need it can add `include 'monitoring/lemon/schema'; ` in their site config and that `monitoring/lemon/schema` should do the type binding.

jrha: @samary the simplest solution here if for you to just delete this line and the one above and add
```pan
bind /system/monitoring = structure_monitoring;
```
to `monitoring/lemon/schema.pan`.

quattor/types/system.pan
```diff
@@ -50,7 +50,7 @@ type structure_system = {
     "lcg"           ? structure_lcg
     "network"       : structure_network
     # TODO: monitoring-related structures should go elsewhere. 
-    "monitoring"    ? structure_monitoring
+    "monitoring"    ? nlist
```
jrha: `dict`

jrha:
@samary I have created a PR on your fork with my proposed changes (samary/template-library-core#1)

jouvin:
Solution proposed with @jrha should work with the template library as it is and the SCDB machine types as the template `site/global_variable` is included before the schema. It's true that defining a variable doesn't a definitive advantage except that is allows to get all the types and units defined and avoid to have to include them in the monitoring schema. My preference would have remained (in the schema):
```
variable MONITORING_SCHEMA_TEMPLATE ?= null;
include MONITORING_SCHEMA_TEMPLATE ;
```
BTW, This solution would also work if the site prefer to just include the schema as suggested by @jrha .

jouvin:
@samary could you merge @jrha PR on your source branch?

jrha:
I gave up waiting and merged both manually.

samary:
@jrha @jouvin : Sorry for not answering in time.
LGTM !
We can close this issue as soon as https://github.com/quattor/template-library-monitoring/pull/4 is merged.
Thanks

PR #127, opened 2016-10-07
types/metadata: add optional id and timestamp to the branch_type

stdweird:


stdweird:
i'm aware that the timestamp and/or id will require a recompile of all profiles.
but the fields are optional, and for us having the id and timestamp of the git branch are more important than the recompilation slowdown

quattor/types/metadata.pan
```diff
@@ -4,6 +4,10 @@ type structure_branch = {
     "author" ? string
     "name" : string
     "type" : string
+    @{a unique id, e.g. hash of last git commit}
+    "id" ? string
```
ned21: id could be a bit mis-leading/ambiguous. What's the name of the pan command option that would fill this in?  How about "profile version" since that's really what this is.

stdweird: i thought about naming this version, but version suggests sometyhing ordered, which a commit hash is not. i can add another optional version field (anf maybe retrict it to `\d+.\d+\.\d+`)

stdweird: it's actually why we use the timestamp also, and not only the hash. the timestamp gives us ordering

stdweird:
using these fields on its own does not lead to the profiles being regenerated, it's how you use them.
e.g. the commit hash only changes on a new commit or a rebase, in which case something actually changed.
same with the timestamp of the last commit.
if you would use the date of compilation however, then sure, every recompile is always a full recompile.

so it's all up to you what you put in it, it's certainly not something the compiler injects. it's similar with adding teh branch name or author.

quattor/types/metadata.pan
ned21: I guess the problem is that this not the id of the branch, as suggested by the namespace. While I see your use case, I am not clear how someone would actually use this: if the commit id is set inserted by the compile then there needs to be something that links that action to this field. 

Having said that, if you are doing some git hook magic to insert this without using a compiler feature, then given this is within the branch subtree, and branches are an artefact of version control, commit-id would probably be a reasonable name.

stdweird: wrt to actual usage, i have no idea how aquilon inserts this metadata, but it's the same way i guess.
we use https://github.com/stdweird/quattor-SCDB-ugent/blob/master/bc#L148 to generate the commanline data and then inject it in the compilation using https://github.com/stdweird/quattor-SCDB-ugent/blob/master/bc#L218
the templates have a dummy default value so one can compile and check, the `bc` script is probably equivalent to `aq deploy` and is used on our deploy hosts.

stdweird:
@ned21 renamed to `commit-id`

quattor/types/metadata.pan
jrha: aquilon inserts the data by writing fields to the host's object template, which works because the broker is in charge of builds.

For example:

```pan
"/metadata/template/branch/name" = "workernode-abc2015";
"/metadata/template/branch/type" = "sandbox";
"/metadata/template/branch/author" = "abc12345";
```

stdweird: which hosts? all hosts? 
i had a short discussion with @piojo about this, we were pondering the possibility of a new pan option to inject data from the commandline/ant task that would not be considered modified data wrt the decision making process that pan uses to determine which object templates to rebuild.
so only the object templates it would rebuild, would get the metadata. but it was over a beer, and certainly not while looking at the pan code :wink:

jrha: Yes, all hosts.

For a host in a domain it would only write something like:
```pan
"/metadata/template/branch/name" = "prod";
"/metadata/template/branch/type" = "domain";
```

Implementing the git commit id would probably be best supported in the compiler IMHO.

PR #128, opened 2016-11-01
Preferred should not be mandatory

apdibbo:
Preferred here should not be mandatory.

I also cannot see anywhere where this value is used


stdweird:
ok to test

jrha:
ok to test

PR #131, opened 2016-11-21
functions/package: add a comment/description to pkg_compare_version funciton

jouvin:
Syntax of the annotation incorrect, I have a meeting, I'll tro fix it when I am done with it.

### DIFF
--- a/quattor/functions/package.pan
+++ b/quattor/functions/package.pan
@@ -1,6 +1,13 @@
 declaration template quattor/functions/package;
 
-# Compare package versions
+@{
+desc = function to compare 2 package versions in the form x.y.z-r. It properly handles \
+ number comparison of each element (10 is considered greater than 2) and it also accepts \
+ letters in one of the element (a string comparison is done for the element in this case).
+args = the two version strings to compare.
+return_value = -1 if the first argument is greater than the first one, 0 if they are equal \
+ and 1 if the second argument is greated than the first.
+}
 function pkg_compare_version = {
     if (ARGC > 1) {
         v1p = ARGV[0];
--- a/quattor/functions/package.pan
+++ b/quattor/functions/package.pan
@@ -1,9 +1,27 @@
 declaration template quattor/functions/package;
 
+# The value for the constants are historical and a bit misleading. Be sure to use the constant names
+# rather than the valueis.
+@{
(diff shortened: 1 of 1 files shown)


quattor/functions/package.pan
```diff
@@ -1,6 +1,13 @@
 declaration template quattor/functions/package;
 
-# Compare package versions
+@{
+desc = function to compare 2 package versions in the form x.y.z-r. It properly handles \
+ number comparison of each element (10 is considered greater than 2) and it also accepts \
+ letters in one of the element (a string comparison is done for the element in this case).
+args = the two version strings to compare.
+return value = -1 if the second argument is greater than the first one, 0 if they are equal \
```
jrha: -1 if the second argument is less than the first, 0 if they are equal

quattor/functions/package.pan
```diff
@@ -1,6 +1,13 @@
...
+ and 1 if the first argument is greated than the second one.
```
jrha: and 1 if the second argument is greater than the first.

jrha:
I was wondering whether we should add something like:

```pan
final variable PKG_VERSION_GREATER = 1;
final variable PKG_VERSION_EQUAL = 0;
final variable PKG_VERSION_LESS = -1;
```

and then update the function to also use the "constants", it might help avoid confusion.

quattor/functions/package.pan
jouvin: Good catch! The last commit should have a valid syntax and the right contents!

jouvin:
You are probably right... Feel free to do it, I am not sure if I will find the time today...

jouvin:
The test failure is due to something wrong with `yumng`...

jouvin:
Done

jrha:
Tests are failing because `SOFTWARE_GROUP` is defined separately for both `yum` and `yumng` and the tests just include both schemas, which should not happen in real usage...

stdweird:
shared types should be reused, not redefined imho. i opened https://github.com/quattor/configuration-modules-core/issues/968 to follow up

stdweird:
retest this please

jrha:
Redefinition is the best way to achieve work around lack of sub-classing, in this case it is a non-issue anyway (see quattor/configuration-modules-core#968).

PR #138, opened 2016-12-14
Support vxvm blockdevices

ned21:
Support for vxvm blockdevices was added via quattor/ncm-lib-blockdevices#60 but
the corresponding template change to blockdevices.pan was overlooked.


stdweird:
@ned21 travis panlint tests fail, but not due to this PR. i'll let @jrha merge this in

jrha:
Thanks!

PR #139, opened 2016-12-14
blockdevices: add validation check configuration to physical devices

stdweird:
Was introduced in https://github.com/quattor/ncm-lib-blockdevices/pull/40

Requires https://github.com/quattor/ncm-lib-blockdevices/pull/73

### DIFF
--- a/quattor/blockdevices.pan
+++ b/quattor/blockdevices.pan
@@ -10,17 +10,18 @@ type blockdev_string = string with exists ("/system/blockdevices/" + SELF)
 
 type physdev_string = string with exists ("/system/blockdevices/physical_devs/" + SELF)
     || error (SELF + " must be a path relative to "
-              + "/system/blockdevices/physical_devs");
+                + "/system/blockdevices/physical_devs");
 
 type parttype_string = string with match (SELF, '^(primary|extended|logical)$');
 
 type vg_string = string with exists ("/system/blockdevices/volume_groups/" + SELF)
     || error (SELF + " must be a path relative to "
-              + "/system/blockdevices/volume_groups");
+                + "/system/blockdevices/volume_groups");
 
 type lv_string = string with exists ("/system/blockdevices/logical_volumes/" + SELF)
     || error (SELF + " must be a path relative to "
-              + "/system/blockdevices/logical_volumes");
+                + "/system/blockdevices/logical_volumes");
+
 @documentation{
     parted partition flags (from "info parted")
 }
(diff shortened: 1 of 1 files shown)


quattor/blockdevices.pan
```diff
@@ -103,10 +103,30 @@ type blockdevices_file_type = {
...
+    @{size-based correctness check}
+    "size" ? blockdevices_correct_size
+};
+
 type blockdevices_disk_type = {
     "device_path" ? raid_device_path
     "label" : string with match (SELF, ("^(none|msdos|gpt|aix|bsd)$"))
     "readahead" ? long # Blocks for readahead
+    @{enable correctness checks}
+    "correct" ? blockdevices_correct
```
ned21: Can we think of a more descriptive name here? "Correct" feels like a boolean, not a dict. How about validate_attributes?

stdweird: how about `@{configure corretcness checks (implies enabling the checks)}`
(i can also change the type name to `blockdevices_correctness`)

ned21: I think the annotation is fine, it's the field name that seemed suboptimal because correct(ness) is an adjective or verb, while a field represents an object (noun).  My ideal suggestion is ```attributes_to_validate``` but that might be a bit long so happy for shorter alternatives.

stdweird: ok, i can't come up with anything; but it also implies a (small) change in `ncm-lib-blockdeivces`.

jrha: Sorry, this is my fault for not identifying this in quattor/ncm-lib-blockdevices#40.

The path when used will appear as `…/correct/size/diff` or `…/correct/size/fraction` which does in fact as @ned21 imply that it will active correct something when it is actually checking a value for correctness.

How about changing it to `tolerance`?

ned21: `tolerance` seems to make sense in the context of `../tolerance/size/diff` and `../tolerance/size/fraction`

stdweird: hmmm, altough tolerance is indeed appropriate for current size checks, i always assumed other cirteria could be added like vendor, serial,... which do not seem good fit for `tolerance`. on the other hand, if the vendor is eg a regexp (eg `'^w`, i.e. a vendor strating with `w`), tolerance might fit.
or maybe `id` / `indentity`  or even `whoami`?

jrha: Hmm, how about `requirement`?

```pan
'requirement/size/diff' = 10;
'requirement/vendor' = 'oceandoor';
```

ned21: This definitely sounds more like `validate`, e.g.

```
'validate/size/diff' = 10;
'validate/size/fraction' = ...;
'validate/vendor' = 'oceandoor';
```

jrha: Works for me.

stdweird: ok, validate it is

stdweird:
@ned21 @jrha please rereview

stdweird:
@ned21 ok, i'll now make the PR for ncm-lib-blockdevices

PR #142, opened 2017-01-13
Add structures to model operating system information

jrha:
We have systems that consume host profiles and need to extract information about the operating system, currently this is done with some nasty regex magic looking at values in locations such as `/system/aii/nbp/pxelinux/kernel` or in one case munging the contents of `/software/components/filecopy/services/{/etc/motd}/config`.

Since the actual information is present in aquilon and during compilation it can be trivially added to the profile.

The reference to OS in the schema at `/system/archetype/os` (defined in quattor/types/aquilon.pan) has been redefined to model the values (name and version) as managed by the aquilon broker.

The current plan is to populate `/system/os` from the `OS_VERSION_PARAMS` as used in template-library-standard, but this could in future be migrated to operating on `/system/os` directly.

Requires #143.

### DIFF
--- a/quattor/types/aquilon.pan
+++ b/quattor/types/aquilon.pan
@@ -41,7 +41,6 @@ type structure_archetype = {
     "name"          : string # e.g. "aquilon"
     @{ Details of operating system as defined by aquilon broker }
     "os"            ? structure_archetype_os
-    "model"         ? string # e.g. "4.0.1-x86_64"
     "filesystem-layout" ? string with if_exists("archetype/filesystem-layouts/" + SELF) != ""
     "archlist"      ? string[] # e.g. fs sysname list for model,
                                # "x86_64.linux.2.6.glibc.2.3", "amd64.linux.2.4.glibc.2.3", ...
--- a/quattor/types/aquilon.pan
+++ b/quattor/types/aquilon.pan
@@ -29,10 +29,18 @@ type structure_cluster = {
     "down_hosts_threshold" ? long(0..)
     "node_index" ? long(0..)
 };
- 
+
+type structure_archetype_os = {
+    @{ Name of OS (i.e. aq --osname or aqdb OperatingSystem.name) }
+    "name" : string
+    @{ Version of OS (i.e. aq --osversion or aqdb OperatingSystem.version) }
+    "version" : string
+};
+
 type structure_archetype = {
     "name"          : string # e.g. "aquilon"
-    "os"            ? string # e.g. "linux"
+    @{ Details of operating system as defined by aquilon broker }
+    "os"            ? structure_archetype_os
     "model"         ? string # e.g. "4.0.1-x86_64"
     "filesystem-layout" ? string with if_exists("archetype/filesystem-layouts/" + SELF) != ""
     "archlist"      ? string[] # e.g. fs sysname list for model,
@@ -115,7 +123,7 @@ type structure_security = {
     "class"         : string with if_exists("archetype/security/" + SELF) != ""
     "svcwhitelist"  ? list
 };
- 
+
 type structure_system_aquilon = {
     "advertise_status" ? boolean
     "archetype"     ? structure_archetype
--- a/pan/types.pan
+++ b/pan/types.pan
@@ -990,3 +990,8 @@ type valid_interface = string with {
     };
     false;
 };
+
+@documentation{
+    desc = CPU architectures understood by Quattor
+}
+type cpu_architecture = string with match (SELF, '^(i386|ia64|x86_64|sparc|aarch64|ppc64(le)?)$');
--- a/quattor/types/annotation.pan
+++ b/quattor/types/annotation.pan
@@ -5,18 +5,18 @@ declaration template quattor/types/annotation;
     optional hardware specific information
 }
 type structure_annotation = {
-   "name"         ? string # "product name"
-   "type"         ? string # "product type"
-   "model"        ? string # "product model"
-   "manufacturer" ? string # "manufacturer name"
-   "vendor"       ? string # "vendor name"
-   "version"      ? string # "product version"
-   "chipset"      ? string # "product chipset"
-   "serialnumber" ? string # "product serialnumber"
-   "arch"         ? string # "product architecture" i386, amd64, m68k, ...
-   "bus"          ? string # "bus of peripheral"
-   "clock"        ? long   # "clock of peripheral"
-   "lang"         ? string # "language of the product"
-   "power"        ? long   # "power in watts"
-   "location"     ? string # "location of the hardware"
+    "name"         ? string # "product name"
+    "type"         ? string # "product type"
+    "model"        ? string # "product model"
+    "manufacturer" ? string # "manufacturer name"
+    "vendor"       ? string # "vendor name"
+    "version"      ? string # "product version"
+    "chipset"      ? string # "product chipset"
+    "serialnumber" ? string # "product serialnumber"
+    "arch"         ? string # "product architecture" i386, amd64, m68k, ...
+    "bus"          ? string # "bus of peripheral"
+    "clock"        ? long   # "clock of peripheral"
+    "lang"         ? string # "language of the product"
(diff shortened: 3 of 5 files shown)


quattor/types/metadata.pan
```diff
@@ -14,7 +14,28 @@ type structure_template = {
...
+    "name" : string
+    @{Major version number component, e.g. 7}
+    "major" : long(0..)
+    @{Minor version number component}
+    "minor" ? long(0..)
+};
+
+type structure_os = {
+    @{Distribution name, e.g. sl}
+    "distribution" : string
```
stdweird: i'm not a big fan of `string` without some basic checking. this is to avoid spelling issues, abreviations etc. there aren't that many we need to add here. i would make a distribution type with official names and also it for the `family` attribute.
is there a reason why it has to be `sl` and not `Scientific Linux`?

jrha: It's supposed to be usable, the definitions of operating systems in our template libraries all use abbreviations. I've add validation to match that in `standard/os/version`.

jrha:
I've split distribution details out to a new type.
I'd like to validate `arch`, ideally with an existing type, but the template library is lacking there.

```pan
 "architecture"  ? string with match (SELF,'i386|ia64|x86_64|sparc')
```

```pan
type systemd_unit_architecture = string with match(SELF, '^(native|x86(-64)?|ppc(64)?(-le)?|ia64|parisc(64)?|s390x?|sparc(64)?)|mips(-le)?|alpha|arm(64)?(-be)?|sh(64)?|m86k|tilegx|cris$');
```

jrha:
Example to populate metadata from `OS_VERSION_PARAMS` (untested):

```pan
prefix '/metadata/os/';
'distribution/name' = OS_VERSION_PARAMS['distribution'];
'distribution/family' = OS_VERSION_PARAMS['family'];
'version/name' = replace('^' + OS_VERSION_PARAMS['distribution'], OS_VERSION_PARAMS['version']);
'version/major' = OS_VERSION_PARAMS['majorversion'];
'version/minor' = OS_VERSION_PARAMS['minor'];
'arch' = OS_VERSION_PARAMS['arch'];
```

quattor/types/metadata.pan
```diff
@@ -14,7 +14,35 @@ type structure_template = {
...
+    @{OS version details}
+    "version" : structure_os_version
+    @{Architecture of OS (may be different to system architecture)}
+    "architecture" : string
+};
+
 type structure_metadata = {
     "template" : structure_template
     "features" ? string[]
+    "os" ? structure_os
```
ned21: The type definitions above are good, and make a lot of sense, but please move this under /system/archetype where it belongs. /metadata is for information about the profile and how it was generated, not the /system.

ned21: Actually, I see your earlier comment about /system/archetype/os already being taken. This just comes directly from the broker and is used to construct the include path for the OS templates, which is why it's just a string. How about /system/archetype/osversion as the top-level structure? Or perhaps better, /system/os ?

stdweird: can we add some annotation to the aquilon schema to describe the metadata a bit better; incl the `structure_metadata` itself (or at least what we should and should add to `/metadata`.

jrha: I'm quite happy to move this under `/system/os`, I don't believe it belongs under `archetype` as it is not archetype specific, I am also concious that you have expressed concern in the past about using `/system` as a dumping ground for information.

I've never been clear what the rules are around `/metadata` in my mind that is what this is (since it is informational), so we definitely should document the philosophy of the schema sections.

`/system/archetype/os` has never contained anything for us, are you sure the broker uses it?

jrha:
I've updated the PR to put the structures under `/system/os` and flipped it to depend on #143 rather than the other way round, this way there is never a commit which adds a new unvalidated type.

jrha:
@hpcugentbot please test this

ned21:
> I'm quite happy to move this under /system/os, I don't believe it belongs under archetype as it is not archetype specific, I am also concious that you have expressed concern in the past about using /system as a dumping ground for information.
>
> I've never been clear what the rules are around /metadata in my mind that is what this is (since it is informational), so we definitely should document the philosophy of the schema sections.

I guess that's a failing of us open sourcing the broker code but not the accompanying templates. I can write a short annotation for /metadata that explains that. Will probably get to it later in the week.
 
> /system/archetype/os has never contained anything for us, are you sure the broker uses it?

ah, right - of course the broker only populates and include in the host plenary. /system/archetype/os will come from your os template, and ours just sets it to the string. In which case, is there any reason not to modify the type of /system/archetype/os ? If you are not using it, then highly unlikely anyone else is and it's redundant. Which brings us to your other point:

> I don't believe it belongs under archetype as it is not archetype specific

In Aquilon, os is per-archetype. We use archetype for classes of host, so a hypervisor, storage appliance and grid node are each their own archetypes with their own set of possibility for OS.

jrha:
If we changed the type of `/system/archetype/os`, I would actually like it to contain the `osname` and `osversion` that match the broker's view of the archetype specific OS record, we have several archetypes that use the same OS, but may have different names for it.

ned21:
@jrha  - makes sense. I meant to look today to see what we actually put in there but I don't think it's relevant. So are proposing ```/system/archetype/os/name``` and ```/system/archetype/osversion``` as well as the fields in your current structure_os ?

jrha:
Yes `/system/archetype/os/name` and `/system/archetype/os/version` were what I was thinking, are you happy if I just replace the existing string?

ned21:
@jrha - yes, please replace the existing string. It leads to some deviation between our definition of archetype and yours, but I think you are moving forward in a sensible way and we should catch up :)

jrha:
@ned21 do you have any insight on the use of `/archetype/model`?
The existing comment is not enlightening.
```pan
    "model"         ? string # e.g. "4.0.1-x86_64"
```

ned21:
@jrha  - sorry, another busy day so I forgot to check our profiles, and I should have looked at the archetype type [definition](https://github.com/quattor/template-library-core/blob/ff6a9e46dba2f07ba6936a807de73d9d8597b6a8/quattor/types/aquilon.pan#L33) before.

We use os for the name (linux, solaris), model for the internal version which has traditionally been of the form of 6.0-x86_64 where 6 is the RH major version number and .0 is our internal release (discrete from RHEL updates which are not represented) and x86_64 is the architecture (intel 64 bit, intel 32 bit or sparc). So I guess "model" is what you have called version? In which case, you could remove model too if the broker does not populate it?

quattor/types/aquilon.pan
```diff
@@ -29,10 +29,18 @@ type structure_cluster = {
     "down_hosts_threshold" ? long(0..)
     "node_index" ? long(0..)
 };
- 
+
+type structure_archetype_os = {
+    @{ Name of OS (OperatingSystem.name) }
+    "name" : string
+    @{ Version of OS (OperatingSystem.version) }
```
ned21: What does OperatingSystem.version mean here? Is this a standard reference to which I am not familiar? Is this a vendor release number or an internal release? Is there any guidance on how to set or use this? e.g. should it match the --osversion field in Aquilon?

jrha: Sorry, yes, it's osversion, it's named OperatingSystem.version in aqdb and aquilon protocols.
I've updated the annotation to hopefully make this clear.

jrha:
@ned21 thanks for the info, I've removed /system/archetype/model.

Our broker doesn't populate anything under `/system/archetype` by itself, it all has to be done by templates.

jrha:
As an aside, we use `/system/siterelease` for our internal OS releases, these are simply an ISO8601 date (e.g. `2017-01-19`).

ned21:
It would be good to move /system/siterelease to /system/archetype/os but that's a separate task.

jrha:
I actually makes sense where it is for us, as we create a single global snapshot for all repositories, so it's independent of anything like archetype or os.
This is nice, because you can compare any two arbitrary hosts for newness.

PR #143, opened 2017-01-13
Add CPU architecture type definition

jrha:
Related to #142 (implements validation of OS architecture).

Tagged as backwards incompatible as this will start to validate previously free-form fields.

### DIFF
--- a/pan/types.pan
+++ b/pan/types.pan
@@ -990,3 +990,8 @@ type valid_interface = string with {
     };
     false;
 };
+
+@documentation{
+    desc = CPU architectures understood by Quattor
+}
+type cpu_architecture = string with match (SELF, '^(i386|ia64|x86_64|sparc|aarch64|ppc64(le)?)$');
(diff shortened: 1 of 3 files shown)


pan/types.pan
```diff
@@ -990,3 +990,8 @@ type valid_interface = string with {
     };
     false;
 };
+
+@documentation{
+    desc = CPU architectures understood by Quattor
+}
+type cpu_architecture = string with match (SELF,'i386|ia64|x86_64|sparc');
```
stdweird: should always be `^(...)$`; or release panc 10.4
also, rhel now supports a lot more arch, incl power and arm

jrha: But does Quattor?

stdweird: it's rhel based, there might be minor issues, but it should work.

jrha:
I've expanded the regex with more 64 bit architectures CentOS builds for.

Issue #149, opened 2017-02-23
remove get_subnet_params from quattor/functions/network


stdweird:
Esp if it is not used anywhere.
It claims to be `This is a 'dirty solution' to work around the lack of bitwise AND in pan... `, which exists in pan.

stdweird:
i grepped our repos, no hits except the definiton itself

jrha:
Yes, but only in our crumbly old SCDB, feel free to remove it.

stdweird:
reintroduced in #154, wa sin use at LAL

PR #150, opened 2017-02-23
quattor/functions/network: cleanup and removal of get_subnet_params function

### LINKED ISSUE #148: quattor/functions/network: boot_nic returns last, not first boot device

stdweird:
Fixes #148 #149
Backwards incompatible due to removal of get_subnet_params function #

### DIFF
--- a/quattor/functions/network.pan
+++ b/quattor/functions/network.pan
@@ -1,206 +1,123 @@
-################################################################################
-# This is 'namespaces/standard/quattor/functions/network.tpl', a pan-templates's file
-################################################################################
-#
-# VERSION:    3.2.9, 25/11/09 16:16
-# AUTHOR:     Martin Bock
-# MAINTAINER: Marco Emilio Poleggi <Marco.Emilio.Poleggi@cern.ch>, German Cancio <German.Cancio.Melia@cern.ch>, Michel Jouvin <jouvin@lal.in2p3.fr>
-# LICENSE:    http://cern.ch/eu-datagrid/license.html
-#
-################################################################################
-# Coding style: emulate <TAB> characters with 4 spaces, thanks!
-################################################################################
-#
-# functions related to "/system/network"
-#
-################################################################################
-
 declaration template quattor/functions/network;
 
 include 'pan/types';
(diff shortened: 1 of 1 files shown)


quattor/functions/network.pan
```diff
@@ -1,206 +1,123 @@
...
+}
+function copy_network_params = {
+    nics = value('/hardware/cards/nic');
 
-    if ( !is_list(subnet_list) ) {
-        error("get_subnet_params first argument must be a list");
+    if (!is_defined(nics)) {
+        error("No network cards defined in the configuration");
+    };
+    if (!is_nlist(nics)) {
```
ned21: It's confusing to refer to this as a dict everywhere and then use is_nlist() in preference to is_dict.

stdweird:
@ned21 ah, and there's even an `is_dict` function (didn't know that, had to look it up 😞 )
@jrha can you add mandatory `is_dict` instead of `is_nlist` to panlint?

stdweird:
@ned21 fixed

jouvin:
I'm late to review this merged PR... I am a bit surprised that it has been merged as it. In particular, I think that as a general rule we should not merge a PR flagged as backward incompatible without a few word in its description summarizing what is the backward incompatibility (done in this PR) and the motivation for this change (missing). In this case, the removal of `get_subnet_params()` is a problem for GRIF configuration... I'll look in more details and open an issue if needed.

stdweird:
it wasn't really clear to me what the code did (it looks complicated for something trivial)
It also had
This is a 'dirty solution' to work around the lack of bitwise AND in pan...
in the comment which is not true. 
I guess we can readd the function if needed, but code should be cleaned up at least.

jouvin:
I think this is really old stuff that I wrote... to make easier to have machines with different network configurations (in terms of subnets, routers...) and retrieve the information based on the machine IP, without adding any specific config. The comment about the lack of bitwise AND in panc was true at that time I remember!
I'll look at reimplement the function in a modernized way...

jrha:
@jouvin see #151.

PR #151, opened 2017-02-23
quattor/functions/network: add ip_in_network function

stdweird:
Requires #150

### DIFF
--- a/quattor/functions/network.pan
+++ b/quattor/functions/network.pan
@@ -121,3 +121,41 @@ function copy_network_params = {
     };
     SELF;
 };
+
+@documentation{
+    descr = Test if ipv4 address is in network range.
+    arg = ipv4 address to test
+    arg = network address (is always recomputed using subnet mask)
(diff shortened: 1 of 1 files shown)


stdweird:
@ned21 rebased

jouvin:
I'm late to review this merged PR... I am a bit surprised that it has been merged as it. In particular, I think that as a general rule we should not merge a PR flagged as backward incompatible without a few word in its description summarizing what is the backward incompatibility and the motivation for this change. In this case, the removal of `get_subnet_params()` is a problem for GRIF configuration... I'll look in more details and open an issue if needed.

stdweird:
The removal was not done here, but in https://github.com/quattor/template-library-core/pull/150

it wasn't really clear to me what the code did (it looks complicated for something trivial)
It also had
`This is a 'dirty solution' to work around the lack of bitwise AND in pan...`
in the comment which is not true. i guess we can readd the function if needed, but code should be cleaned up at least.

PR #153, opened 2017-03-04
blockdevices schema: add support for esp partition flag

### LINKED ISSUE #152: update partition flags
see https://www.gnu.org/software/parted/manual/html_node/set.html for list, need in particular `esp` flag for generic (non-gpt) uefi support

one can use the `boot` flag on  gpt for now.

See also https://github.com/quattor/ncm-lib-blockdevices/issues/81

jouvin:
Fixes #152 .


stdweird:
@jouvin the parted version on centos 7.3 doesn't support `esp` yet, so best to use gpt with boot flag for now in actual templates

PR #156, opened 2017-04-13
introduce boolean to ignore filesystem and/or partition in (ks) AII

stdweird:


ned21:
What's the purpose of this field? When would someone find it useful?

stdweird:
we are trying to use ncm-filesystems on multipath devices that are way too complicated to get working during anaconda (IB SRP targets using named `/dev/mapper/abc` devices).

ned21:
Make sense - can you add that rationale to the commit log please? Or perhaps the annotation? Am feeling inspired/motivated by this [tweet](https://twitter.com/mjg59/status/852633050550648832) today :)

stdweird:
@ned21 commit message updated

PR #162, opened 2017-05-31
annotation: cleanup and add support, procured and site fields

wdpypere:
Cleaned up the documentation and added 3 fields used at UGent

### DIFF
--- a/quattor/types/annotation.pan
+++ b/quattor/types/annotation.pan
@@ -5,18 +5,39 @@ declaration template quattor/types/annotation;
     optional hardware specific information
 }
 type structure_annotation = {
-    "name"         ? string # "product name"
-    "type"         ? string # "product type"
-    "model"        ? string # "product model"
-    "manufacturer" ? string # "manufacturer name"
-    "vendor"       ? string # "vendor name"
-    "version"      ? string # "product version"
-    "chipset"      ? string # "product chipset"
-    "serialnumber" ? string # "product serialnumber"
-    "arch"         ? cpu_architecture # "product architecture" i386, amd64, m68k, ...
-    "bus"          ? string # "bus of peripheral"
-    "clock"        ? long   # "clock of peripheral"
-    "lang"         ? string # "language of the product"
-    "power"        ? long   # "power in watts"
-    "location"     ? string # "location of the hardware"
+    @{product name}
+    "name"         ? string
+    @{product type}
+    "type"         ? string
+    @{product model}
+    "model"        ? string
+    @{manufacturer name}
+    "manufacturer" ? string
+    @{vendor name}
+    "vendor"       ? string
+    @{product version}
+    "version"      ? string
+    @{product chipset}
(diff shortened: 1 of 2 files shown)


stdweird:
@ned21 @gombasg any ideas for a good alternative location for the supprot and procuremnet dates? anything in aquilon metadata perhaps?

quattor/types/annotation.pan
```diff
@@ -5,18 +5,39 @@ declaration template quattor/types/annotation;
...
+    @{clock of peripheral}
+    "clock"        ? long
+    @{language of the product}
+    "lang"         ? string
+    @{power in watts}
+    "power"        ? long
+    @{location of the hardware}
+    "location"     ? string
+    @{campus location of the hardware}
+    "campus"         ? string
```
jrha: We already have `/hardware/sysloc/campus`, `structure_annotation` gets attached to a large number of bits and bobs, we don't this here.

See https://github.com/quattor/template-library-core/blob/master/quattor/types/hardware.pan#L170

quattor/types/annotation.pan
```diff
@@ -5,18 +5,39 @@ declaration template quattor/types/annotation;
...
+    @{Date at which the system support runs. DD-MM-YYYY}
+    "support"      ? string with match(SELF, '^[0-3]\d-[0-1]\d-[1-2]\d{3}$')
+    @{Date at witch the system is procured.}
+    "procured"     ? string with match(SELF, '^[0-3]\d-[0-1]\d-[1-2]\d{3}$')
```
jrha: How about `type_isodate`?

quattor/types/annotation.pan
wdpypere: cool! did not know that. I'll remove campus here.

quattor/types/annotation.pan
wdpypere: Also did not know that, I'll fix it.

jrha: If the support and procured dates apply to the machines, perhaps they should hang off `structure_hardware`? If they are more generally useful then the annotations should reflect that.

wdpypere: OK, I'll move them.

wdpypere:
@jrha remarks addressed

jrha:
@hpcugentbot test this please

quattor/types/hardware.pan
```diff
@@ -208,6 +208,10 @@ type structure_hardware = {
     "sysloc"       ? structure_sysloc
     "nodename"     ? string
     "benchmarks" ? structure_benchmark
+    @{Date at which the hardware support runs out.}
+    "support"      ? type_isodate
+    @{Date at witch the hardware is procured.}
```
jrha: witch → which

wdpypere:
@jrha remark fixed

jrha:
@hpcugentbot test this please

PR #163, opened 2017-10-08
client/rpms: allow to disable ncm-cdispd/cdp-listend installation

jouvin:
- Useful when Quattor is used to manage a VM image rather than a machine (in this case `ncm-ncd` is generally run manually and the profile name doesn't match a host name, making cdp notification useless)

### DIFF
--- a/quattor/client/rpms.pan
+++ b/quattor/client/rpms.pan
@@ -5,21 +5,32 @@
 
 unique template quattor/client/rpms;
 
+@{
+descr = variable to enable installation of ncm-cdispd and ccm-listend
+values = boolean
+default = true
+required = No
+}
+variable QUATTOR_INSTALL_CDISPD ?= true;
+
 include 'quattor/client/version';
 
 '/software/packages' = {
     # python-elementtree is required by YUM on SL5 but not listed as a dependency
     # of any other package
     if ( OS_VERSION_PARAMS['majorversion'] == '5' ) {
-      pkg_repl('python-elementtree');
+        pkg_repl('python-elementtree');
     };
 
     # Quattor
-    pkg_repl('cdp-listend',QUATTOR_PACKAGES_VERSION,'noarch');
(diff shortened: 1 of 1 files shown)


quattor/client/rpms.pan
```diff
@@ -5,21 +5,32 @@
...
+variable QUATTOR_INSTALL_CDISPD ?= true;
+
 include 'quattor/client/version';
 
 '/software/packages' = {
     # python-elementtree is required by YUM on SL5 but not listed as a dependency
     # of any other package
     if ( OS_VERSION_PARAMS['majorversion'] == '5' ) {
-      pkg_repl('python-elementtree');
+        pkg_repl('python-elementtree');
```
ned21: While this might be useful clean up, it's confusing to find a line touched by git blame that then has no explanation in the commit log. Perhaps just leave this whitespace untouched until we have a reason to touch this line? (Or do a one-off global whitespace clean up in the entire file?)

quattor/client/rpms.pan
```diff
@@ -5,21 +5,32 @@
...
 
     # Quattor
-    pkg_repl('cdp-listend',QUATTOR_PACKAGES_VERSION,'noarch');
-    pkg_repl('ncm-cdispd',QUATTOR_PACKAGES_VERSION,'noarch');
     pkg_repl('ncm-ncd',QUATTOR_PACKAGES_VERSION,'noarch');
     pkg_repl('ncm-query',QUATTOR_PACKAGES_VERSION,'noarch');
     pkg_repl('ncm-spma',QUATTOR_PACKAGES_VERSION,'noarch');
 
+    if ( QUATTOR_INSTALL_CDISPD ) {
+        pkg_repl('cdp-listend',QUATTOR_PACKAGES_VERSION,'noarch');
```
ned21: panlint seems to want spaces after the comma

PR #164, opened 2017-10-13
Fix usage of onboot in copy_network_params function

jrha:
The update to the network component made backwards incompatible changes to the schema, in this case changing onboot from a string to a boolean. We should therefore not inject broken values into site configuration.


jrha:
Failure on Jenkins is caused by a different issue.

PR #165, opened 2017-11-22
schema: Add structure for modelling hardware sensors

jrha:
We use this at RAL to configure per-model thresholds for automatic shutdown.

The implementation details of how this data is used is left up to end users, including the interpretation of thresholds, units etc.


apdibbo:
Looks good to me but test failing

jrha:
@apdibbo tests are broken on this repo

PR #168, opened 2017-11-27
Add missing aii schema for opennebula

jrha:
These are provided by the component, but not auto-copied during the release cycle, which has been breaking things.


jrha:
FYI tests will still not pass until https://github.com/quattor/configuration-modules-core/pull/1196 and https://github.com/quattor/configuration-modules-core/pull/1195 are merged and copied back into this repo.

PR #170, opened 2017-12-05
 test_templates: add config values for metaconfig templates 

stdweird:


stdweird:
@jrha only 30 or so left to add 😄

jrha:
This is definitely an improvement, but if I understand this correctly:
1. Only one version will get tested.
2. This will need updating in-sync with changes to `configuration-modules-core` when things change.

stdweird:
@jrha yes, this is definetly not the final solution. probably some discussion during a workshop is needed. the test template (esp for metaconfig) should be maintained in metaconfig as part of a unittest, an djust included here. anyway, that will be for another time...

jrha:
@stdweird I've applied the removal of the reference to the missing version template, could you rebase and squash?

stdweird:
@jrha requires https://github.com/quattor/configuration-modules-grid/pull/133 and templates pushed to master

jrha:
Templates were pulled in during RC1, see https://github.com/quattor/template-library-core/commit/3e42dc500df552ed65893d69983d3f00d09162fa#diff-fa823a08480edde64b6d53b442a783da.

PR #171, opened 2017-12-18
pan types: Add type to model Linux capabilities

jrha:
To be used in code that handles capabilities, e.g quattor/configuration-modules-core#1212.


jrha:
We could ship a list, but then we would have to maintain it in line with kernel changes...

jrha:
Yay :green_apple:!

PR #172, opened 2018-01-31
Allow RAM to be modelled on GPU cards

jrha:
Knowing the amount of RAM that a GPU has is very useful for making deployment and scheduling decisions, so provide the ability for it to be modelled.


jrha:
@stdweird happy to merge?

PR #174, opened 2018-04-19
functions filesystem: partitions_add: fix number of required arguments bug

stdweird:
Code clean-up makes use of pan 10.7 features.

Will need https://github.com/quattor/scdb/pull/46 to be merged to pass Travis tests.

### DIFF
--- a/quattor/functions/filesystem.pan
+++ b/quattor/functions/filesystem.pan
@@ -63,16 +63,12 @@ function lvs_add = {
 # The values in second argument dict  can be either a number interpreted as the partition size or a dict where only
 # the keys 'size' and 'flags' are used (other keys are ignored).
 function partitions_add = {
-    if (length (ARGV) != 2 && length (ARGV) != 3) {
+    if (ARGC < 2 || ARGC > 3) {
         error ("%s: requires 2 or 3 arguments", FUNCTION);
     };
 
     pt = ARGV[1];
-    if (length (ARGV) == 3) {
-        ep = ARGV[2];
-    } else {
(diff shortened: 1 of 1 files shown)


stdweird:
this requires 10.7 compiler

jrha:
Hang on, this *requires* panc 10.7?
If that is the case this makes me very uncomfortable, I thought this just allowed the templates to work with 10.7.

stdweird:
to make it work with 10.7, it's only 2nd commit https://github.com/quattor/template-library-core/pull/174/commits/bc972a7e96500daf1b589d72741be9422f5d3bb1
but the cleanup uses features from 10.7

so, yeah, i should better make a PR that just fixes the `null`/undef` issue

but if panlint fails, and you want me to fix it, i won't unless i can use 10.7....

stdweird:
i made #175 to allow transition to 10.7

jrha:
Sure, thanks!

jrha:
@stdweird can you rebase and fix the conflicts?

quattor/functions/filesystem.pan
```diff
@@ -1,58 +1,53 @@
...
-      return(i);
+    base = "/hardware/harddisks";
+    dsk = value(base);
+    foreach (i; v; dsk) {
+        path = format("%s/%s/boot", base, i);
+        if ( exists(to_string(path)) && value(path) ) {
+            return(i);
+        };
     };
   };
```
jrha: This `};` shouldn't be here.

PR #176, opened 2018-06-03
pan/functions: cleanup and remove support from profile_xxx

jouvin:
- domain_from_object(): remove the required argument for defining the default domain (argument use is now deprecated)
- fullname_from_object(): function deprecated (just return OBJECT)

It's already 10 years than the object template name is required to be the host FQDN! Time for cleanup! In particular, requiring a default domain makes no sense has it will never be used. A deprecation warning is issued if still using the old forms that are normally easy to solve as the call to this function used to be in one place (in SCDB, `global_variables.pan`)

### DIFF
--- a/pan/functions.pan
+++ b/pan/functions.pan
@@ -1,13 +1,4 @@
 ################################################################################
-# This is 'namespaces/standard/pan/functions.tpl', a pan-templates's file
-################################################################################
-#
-# VERSION:    3.2.7, 21/08/09 22:22
-# AUTHOR:     Martin Bock
-# MAINTAINER: Marco Emilio Poleggi <Marco.Emilio.Poleggi@cern.ch>, German Cancio <German.Cancio.Melia@cern.ch>, Michel Jouvin <jouvin@lal.in2p3.fr>
-# LICENSE:    http://cern.ch/eu-datagrid/license.html
-#
-################################################################################
 # Coding style: emulate <TAB> characters with 4 spaces, thanks!
 ################################################################################
 #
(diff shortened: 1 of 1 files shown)


pan/functions.pan
```diff
@@ -92,9 +83,8 @@ function npush = {
     n = dict();
     index = 0;
     while (index < ARGC) {
-        if (!is_string(ARGV[index]))
-        error(format("key is not a string: %s", ARGV[index]));
-        n[ARGV[index]] = ARGV[index + 1];
+        if (!is_string(ARGV[index])) error(format("key is not a string: %s", ARGV[index]));
```
stdweird: not sure what pan compiler version we should be targetting thee days, but as of 10.7, you can remove the format

pan/functions.pan
```diff
@@ -161,33 +151,16 @@ function push_if = {
...
+# from the object template name. Its name is assumed to be the host FQDN.
+#
+#
 ############################################################
 function hostname_from_object = {
     # Check cardinality.
     if (ARGC != 0) error("usage: hostname_from_object()");
 
-    # Check first for the old style profile names.
-    m = matches(OBJECT, "profile_([^\\.]+)(\\.(.*))?");
```
stdweird: shouldn't throw an error instead of just removing the code?

pan/functions.pan
jouvin: IMO, it is a little bit too early to require 10.7 in the core templates...

pan/functions.pan
jouvin: SCDB doesn't support `profile_xxx` for many years and Aquilon supported it. So I don't know how it could even happen... It is enough IMO to thrown an error if the hostname is not a FQDN...

ned21:
@jouvin OK, sorry I mis-understood what this was doing.

PR #177, opened 2018-06-03
Schema fixes required for Aquilon support

jouvin:
As explained in https://github.com/quattor/aquilon/issues/90, current schema is not consistent with plenary templates generated by Aquilon (1.12.58). This PR reconciles them, preserving backward compatibility

This PR also improves the schema layout by factoring out of the base file hardware resources specific to Aquilon. And all the properties mandatory in the Aquilon DB and passed by the broker into the generated templates have been defined as mandatory.

### DIFF
--- a/quattor/types/aquilon/hardware.pan
+++ b/quattor/types/aquilon/hardware.pan
@@ -1,4 +1,4 @@
-# This template extends the base schema with Aquilon-provied resources
+# This template extends the base schema with Aquilon-provided resources
 
 declaration template quattor/types/aquilon/hardware;
 
@@ -30,8 +30,6 @@ type structure_sysloc = {
     "room"       : string
     "bunker"     ? string
     "region"     ? string
-    "dns_search_domains" ? string[]
-    "location"   ? string
 };
 
 # All the resources in this type must be optional for
--- a/quattor/types/aquilon/hardware.pan
+++ b/quattor/types/aquilon/hardware.pan
@@ -1,0 +1,43 @@
+# This template extends the base schema with Aquilon-provied resources
+
+declaration template quattor/types/aquilon/hardware;
+
+include 'pan/types';
+include 'quattor/functions/hardware';
+include 'quattor/types/annotation';
+include 'quattor/types/sensors';
+include 'quattor/physdevices';
+
+@documentation{
+    Rack definition
+}
+type structure_rack = {
+    "column" : string
+    "name"   : string
+    "room"   ? string
+    "row"    : string
+};
+
+@documentation{
+    system location definition
+}
+type structure_sysloc = {
+    "building"   : string
+    "campus"     ? string
+    "city"       : string
+    "country"    : string
+    "continent"  : string
+    "room"       : string
+    "bunker"     ? string
+    "region"     ? string
+    "dns_search_domains" ? string[]
+    "location"   ? string
+};
+
+# All the resources in this type must be optional for
+# the schema to remain usable by non-Aquilon users
+type structure_hardware_aquilon = {
+    "rack"       ? structure_rack
+    "sysloc"     ? structure_sysloc
+};
+
--- a/quattor/types/aquilon/system.pan
+++ b/quattor/types/aquilon/system.pan
@@ -1,4 +1,4 @@
-declaration template quattor/types/aquilon;
+declaration template quattor/types/aquilon/system;
 
 @documentation{
     aquilon-related structures
--- a/quattor/types/aquilon/system.pan
+++ b/quattor/types/aquilon/system.pan
@@ -1,3 +1,5 @@
+# This template extends the base schema with Aquilon-provided resources
+
 declaration template quattor/types/aquilon/system;
 
 @documentation{
@@ -47,8 +49,9 @@ type structure_cluster = {
     "max_hosts" ? long(0..)
 };
 
+@{ Details of operating system as defined by aquilon broker }
 type structure_archetype = {
-    "name"          : string # e.g. "aquilon"
+    "name"          ? string # e.g. "aquilon"
     "os"            : string # e.g. "linux"
     "os_lifecycle"  : string
     "model"         : string # e.g. "4.0.1-x86_64"
@@ -106,7 +109,6 @@ type structure_personality = {
     "description"   ? string
     "class"         ? string with match(SELF, '(INFRASTRUCTURE|APPLICATION)')
     "users"         ? string[]
-    "systemgrn"     ? string[]
     "escalation"    ? string
     "notifyrules"   ? string
     "notifyhours"   ? string
@@ -121,7 +123,7 @@ type structure_personality = {
     # want anything else.
     "maintenance_threshold" ? long(0..100) = 50
     "backups"       ? string
-    "host_environment" : string with match(SELF, "^(dev|qa|uat|prod|infra|legacy)$")
+    "host_environment" : string with match(SELF, "^(dev|qa|uat|prod)$")
     "owner_eon_id"  : long
     "stage"         : string
     "esp"           ? structure_espinfo
--- a/quattor/types/aquilon.pan
+++ b/quattor/types/aquilon.pan
@@ -49,8 +49,8 @@ type structure_cluster = {
 
 type structure_archetype = {
     "name"          : string # e.g. "aquilon"
-    "os"            ? string # e.g. "linux"
-    "model"         ? string # e.g. "4.0.1-x86_64"
+    "os"            : string # e.g. "linux"
+    "model"         : string # e.g. "4.0.1-x86_64"
     "filesystem-layout" ? string with if_exists("archetype/filesystem-layouts/" + SELF) != ""
     "archlist"      ? string[] # e.g. fs sysname list for model,
                                # "x86_64.linux.2.6.glibc.2.3", "amd64.linux.2.4.glibc.2.3", ...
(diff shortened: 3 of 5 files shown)


stdweird:
@jouvin i'm not going to review this, since i'm clueless wrt aquilon

jouvin:
@stdweird but you are not clueless about the schema! Thus my choice to have you in the reviewers in case of something I missed...

jouvin:
@jrha mentioned in https://github.com/quattor/aquilon/issues/90 that he worked on the `archetype/os` with @stdweird and decided to define it as a `dict()` which is fine in principle, except that it is not consistent with what is expected by `ncm-cron`  `config.pan` (which expects a string). Which one should we fix?

jouvin:
@jrha I also agree and probably we should make the schema a little bit more strict, in particular having all the values required by the broker (there are quite a lot) mandatory in the schema. We probably need to complete the discussion on things in the Aquilon DB not  exported by the broker in the object template.

jrha:
The problem of keeping the broker's view of the validation in step with the schema is a problem which still needs solving.

jouvin:
Agreed! But I think this is more a problem of process than a technical issue. And the incentive to address it may be boosted by a larger adoption of Aquilon by the community!

jrha:
There is a technical aspect too, any given version of the broker may be managing builds with any (or many) versions of the template library, so ideally the aquilon types should be provided by the broker itself.

jouvin:
@jrha good point! In the meantime, I added a commit that defines as mandatory all the resources of `aquilon.pan` defined by the Aquilon (1.12.58) broker in the plenary templates. 
This doesn't address the Aquilon-provided fields in `hardware.pan` (country, organisation...). As I don't think they are used by any non Aquilon site, should we factor them out of the main file in the same way this is done for `system.pan` so that we could define as mandatory everything provided by the broker?

jrha:
Yes? It seems like a good idea.

jouvin:
Done!

quattor/types/aquilon/hardware.pan
```diff
@@ -0,0 +1,43 @@
+# This template extends the base schema with Aquilon-provied resources
```
ned21: provied => provided

quattor/types/aquilon/hardware.pan
```diff
@@ -0,0 +1,43 @@
...
+type structure_sysloc = {
+    "building"   : string
+    "campus"     ? string
+    "city"       : string
+    "country"    : string
+    "continent"  : string
+    "room"       : string
+    "bunker"     ? string
+    "region"     ? string
+    "dns_search_domains" ? string[]
```
ned21: I don't remember seeing this in the structure_sysloc I sent you? This wouldn't really belong here I think? Nor location.

quattor/types/aquilon/system.pan
```diff
@@ -55,9 +55,11 @@ type structure_archetype_os = {
 };
 
 type structure_archetype = {
-    "name"          : string # e.g. "aquilon"
+    "name"          ? string # e.g. "aquilon"
     @{ Details of operating system as defined by aquilon broker }
-    "os"            ? structure_archetype_os
+    "os"            : structure_archetype_os
```
ned21: Please revert this to `"os"            : string # e.g. "linux"` since the broker has been generating this prior to the the change that introduced this dict.

quattor/types/aquilon/system.pan
```diff
@@ -55,9 +55,11 @@ type structure_archetype_os = {
 };
 
 type structure_archetype = {
-    "name"          : string # e.g. "aquilon"
+    "name"          ? string # e.g. "aquilon"
```
ned21: should be mandatory to match ours.

quattor/types/aquilon/system.pan
```diff
@@ -55,9 +55,11 @@ type structure_archetype_os = {
...
     @{ Details of operating system as defined by aquilon broker }
-    "os"            ? structure_archetype_os
+    "os"            : structure_archetype_os
+    "os_lifecycle" : string
+    "model" ? string
```
ned21: Should be mandatory to match ours

quattor/types/aquilon/system.pan
```diff
@@ -55,9 +55,11 @@ type structure_archetype_os = {
 };
 
 type structure_archetype = {
-    "name"          : string # e.g. "aquilon"
+    "name"          ? string # e.g. "aquilon"
     @{ Details of operating system as defined by aquilon broker }
-    "os"            ? structure_archetype_os
+    "os"            : structure_archetype_os
+    "os_lifecycle" : string
```
ned21: missing validation: 
`string with match(SELF, "^(evaluation|pre_prod|early_prod|production|pre_decommission|inactive|withdrawn|decommissioned)$")`

quattor/types/aquilon/system.pan
```diff
@@ -112,7 +114,7 @@ type structure_personality = {
     "description"   ? string
     "class"         ? string with match(SELF, '(INFRASTRUCTURE|APPLICATION)')
     "users"         ? string[]
-    "systemgrn"     : string[]
+    "systemgrn"     ? string[]
```
ned21: Recent versions of the broker no longer generate this but I guess you need it for people on old broker versions?

quattor/types/aquilon/system.pan
```diff
@@ -127,9 +129,9 @@ type structure_personality = {
     # want anything else.
     "maintenance_threshold" ? long(0..100) = 50
     "backups"       ? string
-    "host_environment" ? string with match(SELF, "^(dev|qa|uat|prod|infra|legacy)$")
-    "owner_eon_id" ? long
-    "stage"         ? string
+    "host_environment" : string with match(SELF, "^(dev|qa|uat|prod|infra|legacy)$")
+    "owner_eon_id"  : long
+    "stage"         : string
```
ned21: Making some of the mandatory could cause issues for users on an older version of the broker

quattor/types/aquilon/system.pan
jouvin: See the complementary PR https://github.com/quattor/aquilon/pull/94 for the broker...

quattor/types/aquilon/system.pan
jouvin: I agree, it should... But as of Aquilon 1.12.60 it is not added to the hostdata by the broker... Should be fixed IMO... See https://github.com/quattor/aquilon/issues/90

quattor/types/aquilon/system.pan
jouvin: No, I think we should require the last version of the broker for people wanting to use this version of the template library. I think we are only 3 sites running a broker and two of us probably up-to-date! Fixed.

quattor/types/aquilon/system.pan
jouvin: See https://github.com/quattor/template-library-core/pull/177#discussion_r192972142

jouvin:
I think that at this stage this PR is complete and addressing all the @ned21 concerns, except the `archetype/os`. But for this issue, as it is not a changed made by this PR but a one year old change (17.7.0) missed by @ned21, I would prefer that it is not discussed as part of this PR (and doesn't block it) and that an issue is opened to rediscuss the change introduced by https://github.com/quattor/template-library-core/pull/142 if needed.

jouvin:
Rebased on #181... Please check! Not yet tested at LAL...

jouvin:
Tested successfully at LAL with last broker version (1.12.60). The main broker problem not yet addressed is the missing archetype name in the hostdata that prevents marking this property required in the schema. @ned21 do you want the schema to be updated and a broker fix to be submitted or do we stay with the current situation in this PR (my preference) and will change it later once the broker does the appropriate thing...

jrha:
@jouvin my feeling is that the schema should follow the broker.

jouvin:
@jrha agreed. But @ned21 mentioned in one of his comment that he was expecting the archetype `name` to be required which is in fact not possible with the current broker. Thus my remark that we could decide to fix the broker to support the expected schema, even if I think it is not a good practice to create some sort of circular dependencies between issues and I'd prefer this to be addressed later, even if related to the original goal of this PR.

quattor/types/aquilon/hardware.pan
```diff
@@ -0,0 +1,41 @@
...
+    "column" : string
+    "name"   : string
+    "room"   ? string
+    "row"    : string
+};
+
+@documentation{
+    system location definition
+}
+type structure_sysloc = {
```
ned21: In our version all fields of this structure are optional. I think it might be best to stay in sync.

quattor/types/aquilon/system.pan
ned21: OK! In that case then we should also remove infra and legacy options are they are no longer valid.

jouvin:
On the broker-schema coupling, you are right that the tighter the more pain potentially... except if we find a way to remove these templates from the template library and have them distributed/generated by the broker.

The reason for the changes proposed here is to allow sites who want to rely on these properties to know which ones are always present and avoid too much testing in every template. So my preference would be to keep the proposed modifications.

quattor/types/aquilon/hardware.pan
jouvin: Fields changed to mandatory are those who are mandatory at the broker-level... See my [answer](https://github.com/quattor/template-library-core/pull/177#issuecomment-395180163) to your [comment](https://github.com/quattor/template-library-core/pull/177#pullrequestreview-126480852)

jrha:
@jouvin this can be merged if @ned21 is happy now

quattor/types/aquilon/hardware.pan
jrha: @jouvin changing these to mandatory has broken builds with our (admittedly ancient) broker :disappointed:.

PR #180, opened 2018-06-05
Define some simple string types to build upon

jrha:
* Non-empty string
* String without any white-space
* String without leading or trailing white-space

### DIFF
--- a/pan/strings.pan
+++ b/pan/strings.pan
@@ -1,0 +1,34 @@
+@documentation{
+Convenience string types intended for use in building up more complex types.
+}
+declaration template pan/strings;
+
+
+@documentation{
+A string that must contain at least one character (anything other than an empty string).
+}
+type string_non_empty = string(1..);
+
+
+@documentation{
+A string that must contain at least one character and may not contain any whitespace.
+}
+type string_non_whitespace = string_non_empty with !match(SELF, '\s');
+
+
+@documentation{
+A string that must contain at least one character, may contain whitespace, but may not start or end with whitespace.
+}
+type string_trimmed = string_non_empty with match(SELF, '^\S(.*\S)?$');
+
+
+@documentation{
+A string that must contain at least one character and no uppercase characters
+}
+type string_lowercase = string_trimmed with ! match(SELF, '[A-Z]');
+
+
+@documentation{
+A string that must contain at least one character and no lowercase characters
+}
+type string_uppercase = string_trimmed with ! match(SELF, '[a-z]');
--- a/pan/types.pan
+++ b/pan/types.pan
@@ -32,6 +32,7 @@ Data type and function definitions for the following basic types:
 declaration template pan/types;
 
 include 'pan/legacy';
+include 'pan/strings';
 
 @documentation{
 This type implements a date/time format consistent with
--- a/quattor/types/sensors.pan
+++ b/quattor/types/sensors.pan
@@ -1,9 +1,9 @@
 declaration template quattor/types/sensors;
 
 type structure_sensor = {
-    "name" : string with match(SELF, '^.+$')
+    "name" : string_trimmed
     "threshold" : long
-    "unit" : string with match(SELF, '^\S+$')
+    "unit" : string_non_whitespace
 };
 
 type structure_sensor_temperature = {


pan/strings.pan
```diff
@@ -0,0 +1,22 @@
+declaration template pan/strings;
+
+
+@documentation{
+Convenience type specifying a string that must contain at least one character (anything other than an empty string).
+Intended for use in building more complex types.
+}
+type string_non_empty = string with match(SELF, '^.+$');
```
stdweird: `'.'` is enough, you can also test the lenght of the string; that probably even faster

pan/strings.pan
```diff
@@ -0,0 +1,22 @@
...
+
+
+@documentation{
+Convenience type specifying a string that may not contain any whitespace.
+Intended for use in building more complex types.
+}
+type string_non_whitespace = string with match(SELF, '^\S+$');
```
stdweird: use `!match(SELF, '\s')`?

jrha:
Feedback on names welcome, before they become spread all over the place.

pan/strings.pan
```diff
@@ -0,0 +1,22 @@
...
+Intended for use in building more complex types.
+}
+type string_non_empty = string(1..);
+
+
+@documentation{
+Convenience type specifying a string that may not contain any whitespace.
+Intended for use in building more complex types.
+}
+type string_non_whitespace = string with !match(SELF, '\s');
```
stdweird: this allows empty string, not sure if you want that. if it is allowed, maybe clarify in documnetation

pan/strings.pan
```diff
@@ -0,0 +1,22 @@
...
+
+
+@documentation{
+Convenience type specifying a string that may contain whitespace, but may not start or end with whitespace.
```
stdweird: regexp doesn't allow empty string, documentation shoudl reflect that

pan/strings.pan
jrha: Wasn't supposed to be allowed before your suggested change, will fix.

pan/strings.pan
```diff
@@ -0,0 +1,23 @@
...
+}
+type string_non_whitespace = string_non_empty with !match(SELF, '\s');
+
+
+@documentation{
+Convenience type specifying a string that must contain at least one character, may contain whitespace, but may not
+start or end with whitespace.
+Intended for use in building more complex types.
+}
+type string_trimmed = string_non_empty with match(SELF, '^(?=\S).+(?<=\S)$');
```
stdweird: this regexp looks very complex, for something that is probably `^\S(.*\S)?$'

jrha: Probably. I can't spend any more time on this now though.

jrha:
@stdweird updated, I hope the regexs now meet with your approval :suspect:

pan/strings.pan
```diff
@@ -0,0 +1,34 @@
...
+@documentation{
+A string that must contain at least one character (anything other than an empty string).
+}
+type string_non_empty = string(1..);
+
+
+@documentation{
+A string that must contain at least one character and may not contain any whitespace.
+}
+type string_non_whitespace = string_non_empty with !match(SELF, '\s');
```
jouvin: `with match(SELF, '\S')`?

jrha: No, that would mean "at least one non-whitespace character"

jrha:
`\S` → Matches anything other than a space, tab or newline.

pan/strings.pan
jouvin: you are right! I wanted to say: `with match(SELF, '^\S+$')` . But probably yours is more efficient, agreed.

pan/strings.pan
```diff
@@ -0,0 +1,34 @@
...
+A string that must contain at least one character, may contain whitespace, but may not start or end with whitespace.
+}
+type string_trimmed = string_non_empty with match(SELF, '^\S(.*\S)?$');
+
+
+@documentation{
+A string that must contain at least one character and no uppercase characters
+}
+type string_lowercase = string_non_empty with ! match(SELF, '[[:upper:]]');
+
```
jouvin: Base on `string_trimmed` rather than `string_non_empty`?

pan/strings.pan
```diff
@@ -0,0 +1,34 @@
...
+
+@documentation{
+A string that must contain at least one character and no lowercase characters
+}
+type string_uppercase = string_non_empty with ! match(SELF, '[[:lower:]]');
```
jouvin: Same remark as for `string_lowercase`

pan/strings.pan
jrha: Could do, any objections?

jrha:
If you want to be more restrictive, choose from:
`\w` →Matches any letter, digit or underscore. Equivalent to [a-zA-Z0-9_].
`[[:alnum:]]` →POSIX class of the above, potentially allows for Unicode matching.

jouvin:
My point is about using something more restrictive than `^\S+$` for `non_empty` and the others derived from it is that I expect these types to be used in circumstances (e.g. file names) where you don't wnat non printable characters at least. May be `[[:alnum:]]` but I think it is missing `-` at least...

jrha:
`[[:alnum:]-._]`?

jouvin:
About https://github.com/quattor/template-library-core/pull/180#issuecomment-394689678: I don't have an example on the top of my mind but I can imagine it to be useful. Basic idea: no whitespace, trimmed, First character uppercase, others lowercase. But we can add it when we have a use case...

pan/strings.pan
```diff
@@ -0,0 +1,34 @@
...
+@documentation{
+A string that must contain at least one character and may not contain any whitespace.
+}
+type string_non_whitespace = string_non_empty with !match(SELF, '\s');
+
+
+@documentation{
+A string that must contain at least one character, may contain whitespace, but may not start or end with whitespace.
+}
+type string_trimmed = string_non_empty with match(SELF, '^\S(.*\S)?$');
```
ned21: I am not sure about this name. It's important that something that appears in a file far away from where it's defined is obvious to the reader. Does trimming something always mean whitespace? A quick google shows excel and jquery functions that do so maybe it is obvious.

jrha: I'm keen to make sure these names make sense, so very happy to take suggestions.

AFAIK most languages call the function that does this either `trim` or `strip` and `trim` seems to be slightly more common.

jrha: FWIW a quick survey gave me:

Language | Function Name
--|--
.NET | `Trim()`
Arduino | `trim()`
Excel | `trim()`
Go | `Trim()`
Java | `trim()`
JS | `trim()`
PHP | `trim()`
Python  | `strip()`
R | `trimws()`
Ruby | `strip()`

jrha:
@ned21 @stdweird are you both happy with the names?

ned21:
@jrha yes.

PR #181, opened 2018-06-05
Revert /system/archetype/os changes

### LINKED ISSUE #179: Revert os and model fields structure archetype to Aquilon standard
In #142 we tried to remove some of the MS specific terms used for OS versions and implement a more logical structure. This was done on the supposition that:

1.  There was no dependency on those fields from shared parts of the project. This was proven untrue by the discover of the dependency on /system/archetype/name in [ncm-cron](https://github.com/quattor/configuration-modules-core/issues/1281).
1. The structure was populated only by templates and thus individual sites were free to use whatever format they wish. This turned out to be incorrect since the previous October the broker had been changed to populate those fields.

At the time I thought we might be able to adopt the new structure but that subsequently also proven impossible, not only because the broker is now populating those fields which makes it a significant undertaking for us to change, but also because that version field is so important it's consumed as part of our data warehouse code. Combined, these factors mean that changing the field would require a significant coordinated change across multiple code bases and probably downtime so it's not going to happen.

Also subsequently, @jouvin has started to advocate having the broker populate more items in the templates, thus further negating (2) above. In order for this approach to work, the broker and templates **must be** aligned to avoid divergent code paths that are time consuming to maintain and cross-communication issues between the primary broker developers and community users. 

I am cognisant that in encouraging the change in #142 I made a significant error in judgement for which I apologise. I believe that at present only 1 site is effected, @jrha since the three others sites using Aquilon are on more recent versions of Aquilon where the prior structures are now required.

@jrha : how painful would it be for you to back out of this and use an alternative structure?

jrha:
These changes were included as part of #142, but were not necessary for that use-case. The impact of these changes was underestimated and as the new types appear to be unused in the community the change should be reverted ASAP.

Fixes #179.
Fixes quattor/aquilon#93.

### DIFF
--- a/quattor/types/aquilon.pan
+++ b/quattor/types/aquilon.pan
@@ -58,6 +58,7 @@ type structure_archetype = {
     "name"          : string # e.g. "aquilon"
     @{ Details of operating system as defined by aquilon broker }
     "os"            ? structure_archetype_os
+    "model"         ? string # e.g. "4.0.1-x86_64"
     "filesystem-layout" ? string with if_exists("archetype/filesystem-layouts/" + SELF) != ""
     "archlist"      ? string[] # e.g. fs sysname list for model,
                                # "x86_64.linux.2.6.glibc.2.3", "amd64.linux.2.4.glibc.2.3", ...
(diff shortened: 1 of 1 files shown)


jrha:
I'd like to just keep this PR as a reversion, combining changes is what got us into this mess.

jouvin:
I agree. It was my point. We should reverse the "mess" first and rebase https://github.com/quattor/template-library-core/pull/177 on it rather than the opposite...

ned21:
Shall I merge?

PR #184, opened 2018-08-07
Add a valid_user type

XaF:
In some cases, we need to be able to check a username received as
parameter, this allows it by providing a new valid_user_name function
that checks that the username received as parameter is a valid one,
i.e. comprised only of alphanumeric characters, plus a few
special ones: .+:-

### DIFF
--- a/quattor/functions/validation.pan
+++ b/quattor/functions/validation.pan
@@ -139,3 +139,13 @@ function valid_absolute_file_paths = {
     };
     true;
 };
+
+variable VALID_USER_PATTERN ?= '^[a-zA-Z_][a-zA-Z0-9_-]{0,30}([a-zA-Z0-9_-]|\$)?$';
+
+@documentation{
+    This type is meant to be used to check if strings provided are
+    valid user names. This is however designed to be overriden by
+    site-specific policies. The regex used here as default is the
+    one suggested by the useradd man page.
+}
+type valid_user = string with match(SELF, VALID_USER_PATTERN);


quattor/functions/validation.pan
```diff
@@ -139,3 +139,15 @@ function valid_absolute_file_paths = {
     };
     true;
 };
+
+@documentation{
+    desc = Function to validate whether the parameter received is a valid username or not; a valid username is considered to only be comprised of alpha numeric characters, plus a few special ones which are: .+:-
```
stdweird: line too long

quattor/functions/validation.pan
```diff
@@ -139,3 +139,15 @@ function valid_absolute_file_paths = {
...
+}
+function valid_user_name = {
+    user = to_string(ARGV[0]);
+    if(!match(user, "^[A-Za-z0-9._:-]+$")) {
```
stdweird: and the usernames can start with the funky stuff too?

quattor/functions/validation.pan
```diff
@@ -139,3 +139,15 @@ function valid_absolute_file_paths = {
...
+        error(user + " is not a valid username!");
+        return(false);
```
stdweird: this will never be reached. if you want the function to be a proper boolean function, remove the `error` (replace by debug)

stdweird:
i'm not too happy with this, given that the special characters make this an invalid username for us...

quattor/functions/validation.pan
XaF: How do you split `desc` lines ?
Most of the examples I've seen don't split the documentation lines...

quattor/functions/validation.pan
XaF: True, it shouldn't. I actually converted a regex that Morgan Stanley was using internally to a function. Not sure why characters such as `.` or `:` are authorized. Will check. Might change that to the more standard regex for usernames: `^[A-Za-z_]([A-Za-z0-9_-]{0,31}|[A-Za-z0-9_-]{0,30}\$)$`

quattor/functions/validation.pan
XaF: Will the true/false answer be sufficient for using types ?
i.e. `type valid_user = string with valid_user_name;`

quattor/functions/validation.pan
stdweird: usual `\` followed by newline works.
ideally newline and eg some indentation should work too; i opened https://github.com/quattor/pan/issues/210

quattor/functions/validation.pan
stdweird: usernames can start with digits (i think posix allows usernames to be digits only, but lets not go there ;)

quattor/functions/validation.pan
XaF: Ok! Will update. Should I use indentations or not ? (i.e. `\` then writing from the first char of the next line, or using indentations to align such as in the example you give in your opened issue ?)

quattor/functions/validation.pan
stdweird: yes.
do you actually need the function at all? why not make a type to start with, like
```pan
type valid_username = string with match(SELF, someregex);
```
not sure about the error message, if https://github.com/quattor/pan/pull/201 gets merged, the error will contain the failed value

quattor/functions/validation.pan
stdweird: use the indentations, we'll fix the generated docs rather than make the annotation itslef unreadable/ugly

quattor/functions/validation.pan
XaF: If we want to be entirely accurate, we might as well support actually all usernames that can be accepted, from http://manpages.ubuntu.com/manpages/bionic/man8/useradd.8.html (and other resources):

```
       It is usually recommended to only use usernames that begin with a lower case letter or an
       underscore, followed by lower case letters, digits, underscores, or dashes. They can end
       with a dollar sign. In regular expression terms: [a-z_][a-z0-9_-]*[$]?

       On Debian, the only constraints are that usernames must neither start with a dash ('-')
       nor plus ('+') nor tilde ('~') nor contain a colon (':'), a comma (','), or a whitespace
       (space: ' ', end of line: '\n', tabulation: '\t', etc.). Note that using a slash ('/') may
       break the default algorithm for the definition of the user's home directory.

       Usernames may only be up to 32 characters long.
```

The lack of limitation is linked to the fact that PAM allows pretty much anything to be used, while the old passwd was more limitating. Maybe @ned21 would have a better idea as to why this regex is currently used at Morgan Stanley, and what would be acceptable not to break anything for us nor for others ?

quattor/functions/validation.pan
XaF: I don't mind not having it as a function. I'm just starting doing stuff in pan, so most of what I do might be optimized! Once we find a common ground for the regex itself, I'll probably move to using that directly as you suggest. I suggested to merge the type in https://github.com/quattor/configuration-modules-core/pull/1312

stdweird: since 10.7 you can use a type directly to test a value
before that release, people used to make a separate function if they wanted to test for composite types and such. that's why the current template library has some many type functions

XaF: The first occurrence of that code was actually a `if (!match(...))` repeated at multiple places, that I replaced by the above function, that then was suggested to me to use to check the type directly instead of having it in the if. Did not catch at that point that I could do the match directly... héhé.

Regarding the type itself, do you know if there is a better place for me to submit it, or the place I did _is_ the best place? The split of all the templates in multiple repositories is quite confusing for now.

stdweird: @jrha any thougths where to put the type? here or in account component?

quattor/functions/validation.pan
ned21: hmm.. I probably don't know the real history, git blame on the internal repository might yield some light but I doubt it. 

I would go with what is well documented for e.g. ubuntu (can we check RH too?) The one caveat is that I know `.` is definitely required because in some places we ended up using `firstname.lastname` as a userid. I am guessing that also exceeded the 32 character limit that Debian/Ubuntu imposes. 

I would randomly guess that `:` might have been considered valid because in various configuration components, passing `user:group` was a way to set the user and group of a file. In fact, @XaF you will probably find that the there is template code internally that says a field is a "user" but actually accepts `user:group` syntax to allow for the automatic provisioning of directories based on which users may login to a host. (Sorry I forgot about this when discussing the code internally, it was only when comparing the two regexes that I spotted the potential significance of them.)

XaF: From what I understand of the `useradd` man page, the only real "official" limitation is for the first char (Debian/Ubuntu) and for the username length of 32 characters. The rest is only a recommendation. Moreover, as I was saying before, the generalisation of using PAM for authentication made those limitations even blurrier, and we now only have recommendations that apply. Pretty much every system can have a different setup and different rules for usernames (from starting and ending characters to username length).

It seems to be a bigger issue than one that could be tackled in a general way. I feel that if we are very specific about that internally, we might want to keep that type for ourselves and not burden the community with something that will only be useful to us. Thoughts?

ned21: If this validation result in allowing usernames that other sites would consider invalid then I agree: let's defer it to each site to define their own. It would be nice to document that decision somewhere to prevent a re-run of this conversation in the future but I'm not sure where would make sense. This would be a clear use case for being able to override or add to types in PAN: we could then define `type valid_username` with a very basic string type/minimal regex, and comment that this is designed to be overridden by site-specific policies.

quattor/functions/validation.pan
jrha: I guess that user names are used for more than just ncm-accounts, so here is ok.

jrha:
@XaF whose definition of valid user is this?

quattor/functions/validation.pan
XaF: I feel that's a good idea, this would allow to standardize the existence of that variable if we need it in other generic templates, while still keeping it generic enough from the beginning, and site-specific as we recommend to override it.

What would you feel would be a good default value for that? I think that going for the recommended regex defined in the `useradd` man page (that I quoted earlier) might be interesting here. Or would you prefer a regex that allows everything? `^. *$`. Thoughts @ned21 @stdweird @jrha ?

XaF:
@jrha quoted your name in the ongoing conversation about the regex itself. Not sure why github is considering that part of the code outdated.

TLDR; this is Morgan Stanley's regex, but we seem to be moving to using a generic regex and recommending it to be overridden by each site.

XaF:
Updated to set up as type, and added the comment about it being a "default value" meant to be overriden for site-specific uses.

jrha:
@hpcugentbot test this please

jrha:
@stdweird what is your feeling about this now?

If necessary, it seems like it might be easier to override if the pattern were defined as a variable, e.g.
```pan
variable VALID_USER_PATTERN ?= '^[a-zA-Z_][a-zA-Z0-9_-]{0,30}([a-zA-Z0-9_-]|\$)?$';
type valid_user = string with match(SELF, VALID_USER_PATTERN);
```

XaF:
Happy to have it like you suggest @jrha. Just tell me if you want to have it with the variable and I'll update the commit.

jrha:
@hpcugentbot please test this please

stdweird:
@jrha i'm prefer your variable solution, esp if you make it `final`

jrha:
Discussed at workshop, consensus was to use a variable to allow sites to completely redefine the regex.

quattor/functions/validation.pan
```diff
@@ -139,3 +139,11 @@ function valid_absolute_file_paths = {
     };
     true;
 };
+
+@documentation{
+    desc = This types is meant to be used to check if strings provided are \
```
stdweird: remove the `desc = `, and you can also get rid of the `\` at the end of each line

quattor/functions/validation.pan
```diff
@@ -139,3 +139,11 @@ function valid_absolute_file_paths = {
...
+        valid user names. This is however designed to be overriden by \
+	site-specific policies. The regex used here as default is the \
+	one suggested by the useradd man page.
+}
+type valid_user = string with match(SELF, '^[a-zA-Z_][a-zA-Z0-9_-]{0,30}([a-zA-Z0-9_-]|\$)?$');
```
ned21: The problem with this is that validation checks can only be made more-specific in later files, so this would mean that it's impossible for any site to have a regex that's less restrictive (e.g. length >30, or including a .)

```suggestion
variable VALID_USER_PATTERN ?= '^[a-zA-Z_][a-zA-Z0-9_-]{0,30}([a-zA-Z0-9_-]|\$)?$';
type valid_user = string with match(SELF, VALID_USER_PATTERN);
```

quattor/functions/validation.pan
```diff
@@ -139,3 +139,12 @@ function valid_absolute_file_paths = {
     };
     true;
 };
+
+@documentation{
```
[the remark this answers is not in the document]
XaF: Could you please clarify? The variable is already before the type?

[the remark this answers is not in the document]
jrha: I think what @stdweird meant was for you to move the variable before the `@documentation{…` annotation as annotation apply to the thing on the next line.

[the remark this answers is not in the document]
jrha: I've done this for you!

jrha:
@hpcugentbot test this please

jrha:
Travis failure caused by pre-existing lint elsewhere, doesn't need fixing in this PR.

PR #190, opened 2018-10-05
test_templates: Use a single version of schema for ceph and ssh

jrha:
Avoids issues with duplicate type definitions from multi-version schema which have been causing tests to fail on this repository.


jrha:
Merged this myself as it is breaking tests on several other PRs and delaying them unnecessarily.

PR #191, opened 2018-10-25
Allow the caf_service_action to be used by multiple components

jrha:
This will be reverted during the release process unless the corresponding change is made to ncm-metaconfig in configuration-modules-core.

### DIFF
--- a/quattor/types/component.pan
+++ b/quattor/types/component.pan
@@ -2,6 +2,12 @@ declaration template quattor/types/component;
 
 include 'quattor/functions/validation';
 
+type caf_service_actions = string with choice(
+    'restart',
+    'reload',
+    'stop_sleep_start',
+);
+
 type structure_component_dependency = {
     "pre"       ? string[] with is_component_list(SELF)
     "post"      ? string[] with is_component_list(SELF)


quattor/types/component.pan
```diff
@@ -2,6 +2,8 @@ declaration template quattor/types/component;
 
 include 'quattor/functions/validation';
 
+type caf_service_action = string with match(SELF, '^(restart|reload|stop_sleep_start)$');
```
stdweird: use choice?

jrha: Sure, as this is just a move I wasn't going to change the definition, but if you think it's worth it.

components/metaconfig/schema.pan
```diff
@@ -100,8 +100,6 @@ type metaconfig_textrender_convert = {
     true;
 };
 
-type caf_service_action = string with match(SELF, '^(restart|reload|stop_sleep_start)$');
```
stdweird: don't change this, list is not complete

jrha: As discussed at workshop: It seems that ncm-metaconfig artificially restricts which CAF actions are exposed, so for now a more generic solution will require a different type definition.

quattor/types/component.pan
stdweird: this list is also not complete and best to choose otehr name to avoid conflict with metaconfig typ
use eg
```pan
type caf_service_actions = choice('start', 'stop', 'restart', 'reload', 'stop_sleep_start');
```

jrha: Is it really a good idea to allow 'stop' or 'start' as an action to be run when something has changed though?

stdweird:
@jrha https://github.com/quattor/CAF/pull/267

quattor/types/component.pan
```diff
@@ -2,6 +2,12 @@ declaration template quattor/types/component;
 
 include 'quattor/functions/validation';
 
+type caf_service_actions = string with choice(
```
stdweird: i'm going to come back on my own remark:
reintroduce the `caf_service_action` type (and remove ti from metaconfig)
and also introduce the following
```pan
type caf_service_actions = {
    daemons ? caf_service_action{}
};
```
that way we can add support by simply including this type in the schema, rather than copy pasting it everywhere

jrha:
Closed, see discussion in #1335.

PR #193, opened 2018-11-02
Aquilon schema: location attribute requirements relaxed

jouvin:
In previous change, several location-related attributes were made
mandatory as they are required for a machine object. But they
are not for a cluster, with the consequence that they may not be defined
for the virtual machines hosted by the cluster.


jouvin:
Check error is due to `indent.py` called with a wrong option...

PR #195, opened 2019-06-17
Add auto as a valid filesystem type

ned21:
Extend the validation of the type resource in structure_filesystem
to allow "auto" to be used. This is useful for fstab entries which have
been created outside Quattor, and the fstype may not be known.

According to mount(8) the order in which fs types are tried can be
controlled using /etc/filesystems.

### DIFF
--- a/quattor/filesystems.pan
+++ b/quattor/filesystems.pan
@@ -38,7 +38,7 @@ type structure_filesystem = {
     "mountpoint": string
     "mount"     : boolean
     "mountopts" : string = "defaults" # "Mount options"
-    "type"      : string with match (SELF, "^(ext[2-4]|reiserfs|reiser4|xfs|swap|vfat|jfs|ntfs|tmpfs|vxfs|zfs|none)$") # "Filesystem type."
+    "type"      : string with match (SELF, "^(auto|ext[2-4]|reiserfs|reiser4|xfs|swap|vfat|jfs|ntfs|tmpfs|vxfs|zfs|none)$") # "Filesystem type."
     "quota"     ? long # "Quota percentage"
     "freq"      : long = 0 # "Dump frequency"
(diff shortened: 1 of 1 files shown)


ned21:
travis failing due to an indentation issue. To my eyes, the indent looks fine/proper. Is what it accepts documented somewhere so I can compare? I might be happy with what it wants, but I can't tell that unless I know what is expected. 😞

quattor/filesystems.pan
```diff
@@ -38,7 +38,9 @@ type structure_filesystem = {
     "mountpoint": string
     "mount"     : boolean
     "mountopts" : string = "defaults" # "Mount options"
-    "type"      : string with match (SELF, "^(ext[2-4]|reiserfs|reiser4|xfs|swap|vfat|jfs|ntfs|tmpfs|vxfs|zfs|none)$") # "Filesystem type."
+    "type"      : string with match (SELF, "^(auto|ext[2-4]|reiserfs|" +
+                    "reiser4|xfs|swap|vfat|jfs|ntfs|tmpfs|vxfs|zfs|none)$")
+                    # "Filesystem type"
```
jrha: @ned21 I think the indent checker would be happy with the following, but I don't really mind.

```pan
    "type"      : string with match (SELF,
        "^(auto|ext[2-4]|reiserfs|reiser4|xfs|swap|vfat|jfs|ntfs|tmpfs|vxfs|zfs|none)$"
    ) # "Filesystem type"
```

stdweird: or use `choice` and split it like a long list?

ned21:
I see this was the only open PR against 18.12 so I moved it to 19.6. Are we happy to merge as-is rather than doing the more intrusive change to using the choice type?

ned21:
Test this please

PR #196, opened 2020-03-05
pan/types: Add string type for search paths

jrha:
This arose while reviewing quattor/configuration-modules-core#1410.
In an ideal world of course search paths would be modelled as lists of absolute paths, but this should be useful nonetheless.

Based on #197.

### DIFF
--- a/pan/types.pan
+++ b/pan/types.pan
@@ -1022,3 +1022,23 @@ type cpu_architecture = string with match (SELF, '^(i386|ia64|x86_64|sparc|aarch
     desc = Linux capabilities, see CAPABILITIES(7)
 }
 type linux_capability = string with match(SELF, '^CAP_[A-Z_]+$');
+
+@documentation{
+    desc = Search path, a colon seperated set of absolute paths which may include . as a reference to current working directory
+}
+type string_search_path = string_trimmed with {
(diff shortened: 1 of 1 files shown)


pan/types.pan
```diff
@@ -1002,3 +1002,23 @@ type cpu_architecture = string with match (SELF, '^(i386|ia64|x86_64|sparc|aarch
     desc = Linux capabilities, see CAPABILITIES(7)
 }
 type linux_capability = string with match(SELF, '^CAP_[A-Z_]+$');
+
+@documentation{
+    desc = Search path, a colon seperated set of absolute paths which may include . as a reference to current working directory
+}
+type string_search_path = string_trimmed with {
+    if (! match(SELF, '^[^:]+?(?::[^:]+)+$')) {
+        error('The string "%s" does not look like a search path.', SELF);
```
ned21: The error string should include the regex that's being matched against. Otherwise I have to find the source code to figure out why %s doesn't match.

    error(The string "%s" does not look like a search path, i.e. must match /^[^:]+?(?::[^:]+)+$/', SELF);

jrha: Good point!

jrha:
The bugs in the indent check have been fixed upstream and I have opened a separate PR to clean up the rest of the template.

PR #197, opened 2020-03-12
pan/types: Cleanup style - lint, indentation, wrapping

jrha:
Addresses some of the issues flagged by tests (but that were not part of) #196.

### DIFF
--- a/pan/types.pan
+++ b/pan/types.pan
@@ -48,12 +48,15 @@ Ex: 20040825120123Z
 }
 function is_asndate = {
     # Check cardinality and type of argument.
-    if (ARGC != 1 || !is_string(ARGV[0]))
+    if (ARGC != 1 || !is_string(ARGV[0])) {
         error("usage: is_asndate(string)");
+    };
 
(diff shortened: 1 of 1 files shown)


pan/types.pan
```diff
@@ -48,12 +48,13 @@ Ex: 20040825120123Z
 }
 function is_asndate = {
     # Check cardinality and type of argument.
-    if (ARGC != 1 || !is_string(ARGV[0]))
-        error("usage: is_asndate(string)");
+    if (ARGC != 1 || !is_string(ARGV[0])) error("usage: is_asndate(string)");
```
ned21: Really not a fan of this style. 

```
if (...) 
    # action;
```

is much more readable most of the time IMHO, especially when the $condition is contains multiple operators. As a code reader I want to be able to focus on the logic and be able to visually filter out the resulting action.

jrha: Would the following be okay?
```pan
if (...) {
    # action;
};
```

ned21: Even better.

jrha:
I'll rebase #196.

PR #198, opened 2020-03-23
hardware types: add options structure

ned21:
Add a new attribute to the hardware structure to allow the configuration
of optional hardware settings, such as disk encryption.

### DIFF
--- a/quattor/types/hardware.pan
+++ b/quattor/types/hardware.pan
@@ -300,14 +300,14 @@ final variable ENC_TYPES = '^(blade|dumb|hypervisor)$';
 }
 type structure_enclosure = {
     "type"          : string with match(SELF, ENC_TYPES)
-                        || error("enclosure type must be one of: " + ENC_TYPES)
+    || error("enclosure type must be one of: " + ENC_TYPES)
     "children"      ? string[1..] with is_profile_list(SELF)
     "maxchildren"   ? long
 } with {
     if (exists(SELF['children']) && exists(SELF['maxchildren'])) {
         SELF['maxchildren'] >= length(SELF['children'])
         || error("enclosure has too many children, max is " +
-            to_string(SELF['maxchildren']));
+        to_string(SELF['maxchildren']));
     } else {
         true;
     };
--- a/quattor/types/hardware.pan
+++ b/quattor/types/hardware.pan
@@ -244,6 +244,15 @@ type structure_bios = {
 };
 
 @documentation{
+    Configuration settings
+}
+type structure_hw_options = {
+    @{Signal to the hardware management platform that all directly attached
+      disks should be encrypted.}
+    "encrypt_disks" ? boolean
+};
+
+@documentation{
     Hardware definition
 }
 type structure_hardware = {
@@ -258,6 +267,8 @@ type structure_hardware = {
     "sensors" ? structure_sensor_types
     @{Date at which the hardware support runs out.}
     "support"      ? type_isodate
+    @{Optional configuration settings}
+    "options"      ? structure_hw_options
     @{Date at which the hardware is procured.}
     "procured"     ? type_isodate
     # Obsolete field, use the appropriate "cards" sub-field instead!!


ned21:
Is travis having issues? The CI job passes but the message isn't getting back to here? 
https://travis-ci.org/github/quattor/template-library-core/builds/666252097?utm_medium=notification&utm_source=github_status

ned21:
@jrha @stdweird We need to use this internally soon so would it be possible to get some guidance as to whether this is acceptable and likely to eventually be merged please?

jrha:
> Is travis having issues? The CI job passes but the message isn't getting back to here?
> https://travis-ci.org/github/quattor/template-library-core/builds/666252097?utm_medium=notification&utm_source=github_status

Yes, partially degraded at the moment apparently.
https://www.traviscistatus.com/

ned21:
> Yes, partially degraded at the moment apparently.
> https://www.traviscistatus.com/

"The issue only affects those repositories using the legacy Services integration (i.e., all .org repositories and some .com ones), and only under some circumstances. You can migrate your repositories to GitHub Apps on https://travis-ci.com/account/repositories. "

Should we try to move? I'll have a read of the documentation.

> Activate the GitHub Apps integration to start testing and deploying on Travis CI.

> The GitHub Apps integration supports both private and open source repositories, while providing enhanced security when interacting with GitHub.
> 
> We are only able to migrate accounts that have 50 or fewer repositories using the Legacy Services Integration. Please refer to our [documentation](https://docs.travis-ci.com/user/migrate/legacy-services-to-github-apps-migration-guide/) on how to migrate your account.

quattor/types/hardware.pan
```diff
@@ -243,6 +243,13 @@ type structure_bios = {
     "releasedate" : string
 };
 
+@documentation{
+    Configuration settings
+}
+type structure_hw_options = {
+    "encrypt_disks" ? boolean
```
stdweird: what does this mean? encrypt some/all disk?

ned21: Yes, encrypt all disks. Would `encrypt_all_disks` be a better name?

ned21:
travis-ci says it's backup but I can't work out how to refresh the status. It did pass so can I be merged? Or do you want the field name changed?

mochrul:
Hi, I am the original owner of this patch. Is there anything I/we can do to make this patch into the master branch? Is there anything we should change?

We desperately need this functionality, so I would be very thankful for a quick answer.

ned21:
@stdweird @jrha This field name is going to be written to a configuration file and be consumed by an internal tool so it's going to be nearly impossible for us to rename this once we've deployed it. Are you OK for us to use the field name as it is submitted here? We don't need this PR merged, just an "approved" so we know we're not going to be asked to rename it later. 😄

quattor/types/hardware.pan
stdweird: a docstring is enough for me

stdweird:
@ned21 @mochrul sorry this took so long for such seemingly simple change. i don't object a generic options field, but for this particular case, why not a `encrypt ? boolean` for each disk that needs to be encrypted? ( i just looked at the curent disk structures, it looks like a mess tbh; and no easy place to add it...)

quattor/types/hardware.pan
ned21: Done

ned21:
>  but for this particular case, why not a encrypt ? boolean for each disk that needs to be encrypted? ( i just looked at the curent disk structures, it looks like a mess tbh; and no easy place to add it...)

```
quattor/types/hardware.pan:    "harddisks"    ? structure_raidport{}
```

@stdweird - putting this into structure_raidport felt like something that belonged to the OS configuration. i.e. this is where I'd expect to find settings around software/OS based encryption. What we are doing here is generating a vendor-specific config file which can then be processed by whatever tool is managing the hardware RAID controller to do hardware encryption. 

It sounds like we are all good with this so I am going to go ahead and merge @mochrul's change internally.

stdweird:
@ned21 ok, i thought as much. can you extend the docstring a bit more to explain this (and why it's not a per-disk setting)

ned21:
> @ned21 ok, i thought as much. can you extend the docstring a bit more to explain this (and why it's not a per-disk setting)

Done.

jrha:
Closing and re-opening to re-trigger tests.

PR #199, opened 2020-03-26
quattor/functions/hardware: Provide function to get max_threads for all CPUs

jrha:
Much of the configuration of our compute systems relies on knowing both the number of cores and threads available on any given host and as such a number of home-grown "quick fixes" have developed for obtaining this information. This is an attempt to provide the functionality upstream with the hope of removing out local kludges in future.

### DIFF
--- a/quattor/functions/hardware.pan
+++ b/quattor/functions/hardware.pan
@@ -1,4 +1,3 @@
-
 declaration template quattor/functions/hardware;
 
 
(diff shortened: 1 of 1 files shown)


quattor/functions/hardware.pan
```diff
@@ -1,39 +1,92 @@
...
+        if ( is_dict(ARGV[0]) ) {
+            hw_config = ARGV[0];
+        } else if (is_string(ARGV[0])) {
+            hw_config = create(ARGV[0]);
+        } else {
+            error('Invalid argument type (%s)', to_string(ARGV[0]));
+        }
     } else {
-      error(format('get_num_of_cores requires 0 or 1 argument (%s specified)',ARGC));
+        error('get_num_of_cores requires 0 or 1 argument (%s specified)', ARGC);
```
ned21: Use FUNCTION here to avoid the typo in the error message?  (This is no longer get_num_of_cores)

quattor/functions/hardware.pan
```diff
@@ -1,39 +1,92 @@
...
+        debug('%s: num of CPUs=%d, num of cores=%d', OBJECT, length(hw_config['cpu']), cores);
+    } else {
+        error('Invalid hardware configuration (no CPU defined)');
+    };
+    result;
+};
+
+
+@documentation{
+    descr = Returns the number of cores on the current machine or another machine
```
ned21:
```suggestion
    descr = Returns the number of cores for the supplied hardware config, or the current machine if no argument is passed.
```

quattor/functions/hardware.pan
```diff
@@ -1,39 +1,92 @@
...
     };
 
-    cores;
+    d = get_num_of_cores_max_threads(hw_config);
+    d['cores'];
 };
 
+
+@documentation{
+    descr = Returns the maximum number of hardware threads on the current machine or another machine
```
ned21: Same suggestion as above.

jrha:
@ned21 thanks for the feedback! Updated as per your suggestions.

PR #200, opened 2020-06-08
Add string type for an octal mode/mask

wpoely86:
As suggested by @gombasg, just parse it numerically to fit in the interval [0,07777].

@ned21 as requested in quattor/configuration-modules-core#1396

### DIFF
--- a/pan/types.pan
+++ b/pan/types.pan
@@ -1042,3 +1042,9 @@ type string_search_path = string_trimmed with {
     if (dot_count > 1) error('The search path "%s" contains more than one reference to the working directory.', SELF);
     true;
 };
+
+
+@documentation{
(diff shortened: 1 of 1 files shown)


pan/types.pan
```diff
@@ -1042,3 +1042,18 @@ type string_search_path = string_trimmed with {
     if (dot_count > 1) error('The search path "%s" contains more than one reference to the working directory.', SELF);
     true;
 };
+
+
+@documentation{
+Returns true if the string is a valid octal mode/mask
+}
+function is_octal_mode = {
```
stdweird, line 1050: no need for a function imho. simply make a type with the regexp.

wpoely86: done

stdweird:
@wpoely86 thanks!

PR #201, opened 2020-06-12
add caf_serviceaction type

stdweird:


stdweird:
there's already a `caf_service_action` in metaconfig, but this should really be the list from `CAF::ServiceActions` (and is needed for eg https://github.com/quattor/configuration-modules-core/pull/1341 or to add daemon support to ncm-download)

PR #203, opened 2020-11-13
Add NVMe to block device interface types

jrha:
Also switch to `choice()`


jrha:
@ned21 tests pass now :smile:

ned21 (approved):

PR #206, opened 2021-04-28
Use github actions for tests

wpoely86:
the title says it all.

### DIFF
--- a/components/ceph/config.pan
+++ b/components/ceph/config.pan
@@ -16,7 +16,7 @@
 
 unique template components/ceph/config;
 
-variable CEPH_SCHEMA_VERSION ?= 'v1';
+variable CEPH_SCHEMA_VERSION ?= 'v2';
 
 include if_exists('components/ceph/site-config');
 include format('components/ceph/%s/schema', CEPH_SCHEMA_VERSION);
--- a/test_templates.sh
+++ b/test_templates.sh
@@ -395,6 +395,9 @@ sed -i '/.*metaconfig\/\(elasticsearch\|logstash\|beats\)\/.*_[0-9].*/d' build_t
 # none-versioned ganesha templates are v1
 sed -i "/.*metaconfig\/ganesha\/\(config\(_v1\)\?\|schema\)'/d" build_temp/test.pan
 sed -i "/.*metaconfig\/ganesha\/fsal.*/d" build_temp/test.pan
+# action.pan is included by the schema and can only be included after inputs.pan
+sed -i "/.*metaconfig\/rsyslog\/actions.*/d" build_temp/test.pan
+
 # only spma yum
 sed -i "/.*components\/spma\/\(apt\|ips\).*/d" build_temp/test.pan
 
--- a/test_templates.sh
+++ b/test_templates.sh
@@ -391,7 +391,7 @@ find . -type f ! -regex '^./pan/.*' ! -regex '^./build_temp/.*' -name *.pan | gr
 find . -type f ! -regex '^./build_temp/.*' -name *.pan | grep -v yumng | xargs sed -n "s/^\(unique[ ]\+\)\?template[ ]\+\(.*\);/include '\2';/p" |sort >> build_temp/test.pan
 
 # fix for multiversion metaconfig
-sed -i '/.*metaconfig\/\(elasticsearch\|logstash\)\/.*_[0-9].*/d' build_temp/test.pan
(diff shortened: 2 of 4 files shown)


wpoely86:
Output of the actions can be found here: https://github.com/wpoely86/template-library-core/actions

The template one actually fails with:
```
evaluation error [/__w/template-library-core/template-library-core/./metaconfig/beats/schema_6.0.pan:6.1-6.102]
duplicate type definition: beats_output_logstash_ssl_protocol
previous definition [/__w/template-library-core/template-library-core/./metaconfig/beats/schema_pre6.0.pan:6.43-6.53]

>>> call stack trace 
>>> [/__w/template-library-core/template-library-core/./metaconfig/beats/schema_6.0.pan:6.1-6.102]
>>> [/__w/template-library-core/template-library-core/build_temp/test.pan:453.1-453.38]
```

ned21 (dismissed):

ned21:
> The template one actually fails with:

Is that why Jenkins is failing? Looks like it might have been introduced by an automated commit which imports the templates from the components repo?

wpoely86:
Those files came with 056c93c567bc3f9c9151e1ad741ca78a8ee2fb36. The issue is that you have 3 files:
- `schema_6.0.pan`
- `schema_6.3.pan`
- `schema_pre6.0.pan`

All which define `beats_output_logstash_ssl_protocol`. Normally only one should be included but this test takes all of them?

ned21:
Yep. I think https://github.com/quattor/template-library-core/blob/master/test_templates.sh#L393 needs an update. It looks like it's already trying to exclude logstash?

ned21:
@hpcugentbot Test this please

wpoely86:
I guess it's a while since these jenkins tests ran?

```
evaluation error [/__w/template-library-core/template-library-core/./metaconfig/rsyslog/actions.pan:99.1-119.2]
referenced type 'rsyslog_input_common' is not defined
```

stdweird:
well, these tests try to make a profile with as much templates as possible included. but that is a manual process to get the total list correct.

wpoely86:
So what's the proper way of fixing this? I think the issue is that it first includes `actions.pan` before `inputs.pan` so it is missing `rsyslog_input_common`

ned21:
metaconfig/rsyslog/schema.pan includes the other files so I think it would be OK to take the same approach as for sshd (line 405) and delete any include line in metaconfig/rsyslog/* that does not match schema? Or perhaps easier, just delete the input, actions and common lines?

wpoely86:
OK, I got it to pass: https://github.com/wpoely86/template-library-core/actions

with ceph the default is `v1` but the templates for v1 are no longer there?

test_templates.sh
```diff
@@ -404,6 +404,9 @@ sed -i "/.*components\/ceph\/v[1-9][0-9]*\/.*/d" build_temp/test.pan
 # Only test default schema version of ssh component
 sed -i "/.*components\/ssh\/schema-.*/d" build_temp/test.pan
 
+# action.pan is included by the schema and can only be included after inputs.pan
```
ned21, line 407: nitpick: can you put this at line 398 so it's co-located with the other metaconfig fix ups please?

components/ceph/config.pan
```diff
@@ -16,7 +16,7 @@
 
 unique template components/ceph/config;
 
-variable CEPH_SCHEMA_VERSION ?= 'v1';
+variable CEPH_SCHEMA_VERSION ?= 'v2';
```
ned21, line 19: Not a blocker but I think needs to get fixed here too or I think it will get reverted at next release: https://github.com/quattor/configuration-modules-core/blob/400aba5c9c5c344fc8b88f0a54209d038fbaedaa/ncm-ceph/src/main/pan/components/ceph/config.pan

ned21:
@hpcugentbot test this please

components/ceph/config.pan
wpoely86: https://github.com/quattor/configuration-modules-core/pull/1506

test_templates.sh
wpoely86: done

ned21:
Thanks for quick turnaround. Can you rebase -i and squash 1acb75c58c783aaf6d0805a54a69f6d8507d4f49 into c68afefc69ad7dcfe2165644862f76a647f41a53 please to keep the history easy to follow?

ned21:
@hpcugentbot test this please

ned21 (approved):

ned21:
@wpoely86 Thank you for the awesome work here - it's much appreciated. 🏆 🍻

PR #207, opened 2021-05-10
add support for proc filesystem (and dummy proc blockdevice)

stdweird:


quattor/blockdevices.pan
```diff
@@ -155,12 +155,20 @@ type blockdevices_vxvm_type = {
...
 @documentation{
     TMPFS devices (dummy devices)
 }
 type blockdevices_tmpfs_type = {
 };
 
+@documentation{
+    Dummy devices (e.g. tmpfs, proc)
+}
+type blockdevices_dummy_type = {
```
ned21, line 168: Maybe, s/dummy/empty/ here? Or "special" ? Or placeholder ?

My reasoning is that these still have an fs_spec, it's just the way that Quattor works that requires them to have an entry in structure_blockdevices that forces us to have this type? So to call them of type "dummy" is slightly misleading because we don't actually create anything, this is all quattor internal plumbing.

quattor/filesystems.pan
```diff
@@ -39,7 +39,7 @@ type structure_filesystem = {
     "mount"     : boolean
     "mountopts" : string = "defaults" # "Mount options"
     "type"      : string with match (SELF, "^(auto|ext[2-4]|reiserfs|" +
-                    "reiser4|xfs|swap|vfat|jfs|ntfs|tmpfs|vxfs|zfs|none)$")
+                    "reiser4|xfs|swap|vfat|jfs|ntfs|tmpfs|vxfs|zfs|none|proc)$")
```
ned21, line 42: Is it time to re-order this into some sort of logical order? e.g. 
- auto
- none
- proc
- swap
- ext[2-4]
- jfs
- ntfs
- reiser4
- reiserfs
- vfat
- xfs
- zfs

ned21 (commented):
Any idea what's wrong with the indent script?

quattor/blockdevices.pan
ned21: In the context of https://github.com/quattor/ncm-lib-blockdevices/pull/94#pullrequestreview-655827634 then maybe this is the right term.

PR #209, opened 2021-10-27
El8 support

jouvin:
- Add the missing template for AII KS
- Ensure that the `ncm-spma` version used is the one configured, if the default version was changed (.e.g. for testing)


jouvin:
I checked with `pan_indent_checker` too in the source branch of this PR and cannot see any problem:

```
git diff --name-only HEAD^ | grep '\.pan$' | xargs -r /exp/si/jouvin/GitRepositories/Quattor/Pan-Indenter/pan_indent_checker.py  check
```

Not sure why the test crashed and I don't see how to rerun it...

jrha (approved):

PR #211, opened 2022-03-10
AII KS: use the base OS repository template to configure the AII urls

jouvin:


stdweird:
@jouvin not sure how this relates to https://github.com/quattor/aii/pull/316

jouvin:
@stdweird I think you are right that they try to address the same issue. I don't see any conflict between them. Mine builds the appropriate URL at the template level, the other PR is adding a new URL syntax with a placeholder parsed by the component. I see this approach based on the component has never been merged since 2 years, not sure why. Mine is backward compatible (if you don't define the variable it will do the same thing as before) so I'd suggest to merge it as it is a show stopper to deploy a server with AII otherwise.

jouvin:
The `panlint` test is again failing to run in the context of the CI for reasons I don't understand. But running locally, it reports no error with any of the modified file.

jouvin:
@stdweird @jrha I see that this PR was never merged, I don't understand why... Any reason not to merge it?

jrha (approved):

PR #212, opened 2022-04-08
Add a transitional type for strings being migrated to lists of strings

jrha:
This is aimed at solving the migration path for quattor/configuration-modules-core#1464 in a backwards compatible way.


ned21 (approved):

wpoely86:
Looks good to me. Can we merge this? I will adjust quattor/configuration-modules-core#1464

PR #214, opened 2022-10-14
New function lvo_add to add lv options

aka7:
New function, lvo_add. This is to replace lvs_add function which only sets size when adding of logical volumes to a volume group. with this function, you can provide list of lv options to add, raid_type is one of them. Creating raid through LVM is now possible, therefore this function provides support for setting the type for logical volumes.

fixes https://github.com/quattor/template-library-standard/issues/150

### DIFF
--- a/quattor/functions/filesystem.pan
+++ b/quattor/functions/filesystem.pan
@@ -50,6 +50,34 @@ function lvs_add = {
 };
 
 @documentation{
+    desc = Adds a list of logical volumes to a volume group with mutiple lv options.
+    arg = There are 2 mandatory arguments:\
+        1: string - the volume group name\
+        2: list of dicts representing the LVs desired. Supported keys within each dict are: size, raid_type
+}
+# deprecates lvs_add function.
+function lvo_add = {
+    if (ARGC != 2) {
+        error ("%s: should get 2 arguments", FUNCTION);
+    };
+    vg = ARGV[0];
+    lv_list = ARGV[1];
+    # for each logical volume options defined, set the values.
+    # Only size and raid type is supported.
+    foreach (i; opt; lv_list) {
+        SELF[i]["volume_group"] = vg;
+        if (is_defined(opt['size']) && opt['size'] != -1) {
+            SELF[i]["size"] = opt['size'];
+        };
+        if (is_defined(opt['raid_type'])) {
+            SELF[i]["type"] = opt['raid_type'];
+        };
+    };
+    # Validation stuff might be added here.
+    SELF;
+};
+
+@documentation{
     desc = Adds a list of partitions to a disk
     arg = holding device, list of partitions to creatd, optionally id of the extended partition,
 }


aka7:
My first thought was to just modify lvs_add function,  but decided against it to avoid any compatibility issues. not sure if there will be one since filesysystem/config.pan is also being modified.
 so happy to hear thoughts on this from quattor experts.

quattor/functions/filesystem.pan
```diff
@@ -49,6 +49,32 @@ function lvs_add = {
     SELF;
 };
 
+@documentation{
+    desc = Adds a list of logical volumes to a volume group with mutiple lv options. To replace function lvs_add which only sets size.
+    arg = volume group, list of logical groups long with optons.  size and raid type are set if defined.
+}
+function lvo_add = {
+    function_name = 'lvo_add';
```
ned21, line 57: You can drop this line, I think it's historical. The FUNCTION global is the new way to do this.

quattor/functions/filesystem.pan
```diff
@@ -49,6 +49,32 @@ function lvs_add = {
     SELF;
 };
 
+@documentation{
+    desc = Adds a list of logical volumes to a volume group with mutiple lv options. To replace function lvs_add which only sets size.
```
ned21, line 53: Sorry if this is obvious but is the implication should this deprecate lvs_add and people encouraged to move over? (I see your comment and agree this is for discussion)

quattor/functions/filesystem.pan
```diff
@@ -49,6 +49,32 @@ function lvs_add = {
...
+    arg = volume group, list of logical groups long with optons.  size and raid type are set if defined.
```
ned21, line 54: I know this is following the convention of this file but I think this can be made clearer. There are probably more modern examples of how to document functions nicely elsewhere in the code base which maybe someone can point at but if not here's my suggestion:

```
    There are 2 mandatory arguments:
        1: string - the volume group name
        2: list of dicts representing the LVs desired. Supported keys within each dict are: size, raid_type
```

quattor/functions/filesystem.pan
```diff
@@ -49,6 +49,32 @@ function lvs_add = {
...
+}
+function lvo_add = {
+    function_name = 'lvo_add';
+    if (length (ARGV) != 2) {
+        error ("%s: should get 2 arguments", FUNCTION);
+    };
+    vg = ARGV[0];
+    lv = ARGV[1];
```
ned21, line 62: This is a list so it should be lvs (which is horrible) so `l_vols` ? or `lv_list` ?

quattor/functions/filesystem.pan
```diff
@@ -49,6 +49,32 @@ function lvs_add = {
...
+    # for each logical volume options defined, set the values.
+    # Only size and raid type is supported.
+    foreach (l; opt; lv) {
```
ned21, line 65: lv is a list so make it clear l is an index:

`foreach (i; lv; lv_list)`

ned21 (commented):
I see what you mean about it being difficult to modify `lvs_add`. I think you could do it by allowing ARGV[1] to be either a list of longs or a list of dicts. Then at line 44 you do:

```
if (is_long(sz)) { 
   # old behaviour
} else {
   # new behaviour
};
```

But the original code uses some odd variables names (e.g. l for an index instead of i) so just starting over and deleting the old function in due course might be a better idea.

quattor/functions/filesystem.pan
aka7: thanks, I couldn't find an example of Modern way of doing this.  It appears  arg =  seems want all in one line, so added your suggestion by escaping it.

quattor/functions/filesystem.pan
aka7: Not sure how the two releases of template standard and core are released, whether they are released together? If they are, then this PR will solve any compatibility issue, even if lvs_add is modified. 
https://github.com/quattor/template-library-standard/pull/151

so no need to modify line 987 in above pr if lvs_add is modified to take list of dict as 2nd arg instead of long value

stdweird: i would add the comment of relplacing lvs_add as a comment in the pan code, not part of the docs. as you mention, it's a one time thing to do

aka7: ok, make sense. done.

jrha:
Syntax/Indent check is currently failing on pre-existing code, so safe to ignore.

jrha (approved):
LGTM

Issue #219, opened 2023-07-03
schema ntpd_tinker_options/step should be "double", not "long"


ulrich1919:
See tinker/step definition in https://www.ntp.org/documentation/4.2.8-series/miscopt/

PR #220, opened 2023-07-03
Fix type of ntpd_tinker_options/step #219

ulrich1919:
The step value is represented in seconds.
See tinker definition in https://www.ntp.org/documentation/4.2.8-series/miscopt/


ulrich1919:
see also https://github.com/quattor/configuration-modules-core/pull/1599

jrha:
Lint is in existing (non-modified) code, so that failure can be ignored.

jrha:
LGTM, template should be updated from the component during the release process.

jrha (approved):

PR #221, opened 2023-12-14
add bootmode option to provide support for uefi

aka7:
bootmode with a choice of "uefi" or "bios" added to the bios subtree


jrha:
I don't have a problem with this in principle, but do you have a way of enforcing it?
Right now RAL and other sites configure bootservers to support both Legacy BIOS and UEFI in parallel.

jrha:
Fix for tests is in #222.

PR #230, opened 2024-11-04
AII KS variant for EL9: fix required packages and Anaconda version

jouvin:
Changes that were needed in the EL9 variant at IJCLab...


jouvin:
Tests failing because of the syntax problem in metaconfig/beats, fixed in `configuration-modules-core` but probably not imported yet in `template-library-core`

jrha (approved):

PR #232, opened 2024-11-06
test-templates: Don't include network backend specific schema

jrha:


jrha:
The remaining failures are due to the AII release being missing.

PR #235, opened 2024-11-12
quattor/functions/network copy_network_params: bootproto=none for unconfigured interfaces

jouvin:
- Required for `nmstate` to work properly for unconfigured interfaces

I mark this PR as backward incompatible as it changes the resulting profile compared to previous version but I think it fixes something that was wrong but working with initscripts (because they don't care about interfaces not enabled on boot).

### DIFF
--- a/quattor/functions/network.pan
+++ b/quattor/functions/network.pan
@@ -57,14 +57,15 @@ function boot_nic = {
     default = none
     required = no
 }
-variable MTU ?= dict();
+variable INTERFACE_MTU ?= dict();
 
 @documentation{
     descr = This function configures all the network interfaces declared in\
             /hardware/cards/nic. Parameters are taken from variable\
             NETWORK_PARAMS (dict) for the main (boot) interface, others\
-            are configured with dhcp. For every interface, if there is\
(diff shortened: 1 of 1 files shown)


jouvin:
The `panlint` error is due to a variable called `MTU`, a too short name. I'd prefer to keep it to avoid breaking backward compatibility but if you prefer to do it, it should not be too much an issue for GRIF... If we keep it, is there a magic line to avoid checking this name and raising an error?

jrha:
I'd prefer if it were called something less ambiguous like `INTERFACE_MTU`, I've checked and we don't use it at RAL.

jrha:
I don't think I've shared it before, but we use this simple script for situations where we need to rename something globally: https://gist.github.com/jrha/f8373418cac4ed2ad6f5e5be08d096c3

jouvin:
Ok, I'll do it.

jrha:
Thanks!

jrha:
Can you remove the `(Ã)` from the commit message?

jouvin:
Sure!

jouvin:
Should be ok now...

PR #237, opened 2024-12-08
quattor/functions/network get_subnet_params: allow to return subnet

jouvin:
By default, `get_subnet_params()` return value is compatible with the interface schema (`/system/network/interfaces`). Thus the subnet parameter which is part of the `get_subnet_params()` input is removed. Sometimes, it may be desirable to get it returned. To make it possible, a third optional argument has been added (boolean). When `true`, the subnet is return. Default is unchanged (subnet not returned to be compatible with the interface schema).

### DIFF
--- a/quattor/functions/network.pan
+++ b/quattor/functions/network.pan
@@ -173,14 +173,21 @@ function ip_in_network = {
  by structure_interface + 'subnet' which is the subnet number (result of applying the\
  mask to the address. `subnet` key is mandatory, 'netmask` defaults to 255.255.255.0.
     arg = ip address to test
+    arg = boolean. When true, keep subnet parameter in return value. Default: false.
 }
 function get_subnet_params = {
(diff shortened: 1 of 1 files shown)


jouvin:
`test-library` error is due to the master branch renaming... I need to fix `get_template_library`.

jrha (approved):

jouvin:
I need to fix the problem with `test-library`. Not sure what the cause is but I guess it is somewhat related to the changes done in `template-library-grid` and `template-library-os` architecture or the `get-template-library` related changes...

jrha (approved):

jouvin:
See https://github.com/quattor/template-library-examples/pull/42 for the details on the errors here. In summary:

- `template-libary-examples` fix must be merged
- https://github.com/quattor/configuration-modules-core/issues/1787 must be fixed

PR #238, opened 2024-12-16
aii/server/config: configure grub2_efi_kernel_root

jouvin:
Upstream of this change forgotten before 24.10 release...

PR #239, opened 2024-12-16
quattor/server/config: enable tftpd service on EL9

jouvin:
This change should have been upstreamed before 24.10 release. Required for a working AII server on EL9.

### DIFF
--- a/quattor/aii/server/config.pan
+++ b/quattor/aii/server/config.pan
@@ -11,8 +11,23 @@ variable AII_DHCP_DHCPD_CONF_FILE ?= '/etc/dhcpd.conf';
 
 include 'components/aiiserver/config';
 
-# aii-shellfe configuration
+# Enable tftpd (on EL9 only for backward compatibility)
+include 'components/systemd/config';
+'/software/components/systemd/unit/tftp' = {
+    if ( OS_VERSION_PARAMS['majorversion'] >= '9' ) {
+        SELF['type'] = 'socket';
+        SELF['state'] = 'enabled';
+        SELF['startstop'] = true;
+    };
 
+    if ( is_defined(SELF) ) {
+        SELF;
+    } else {
+        null;
+    };
+};
+
+# aii-shellfe configuration
 
 prefix '/software/components/aiiserver/aii-shellfe';
 


quattor/aii/server/config.pan
```diff
@@ -11,8 +11,18 @@ variable AII_DHCP_DHCPD_CONF_FILE ?= '/etc/dhcpd.conf';
...
-# aii-shellfe configuration
+# Enable tftpd (on EL9 only for backward compatibility)
+include 'components/systemd/config';
+'/software/components/systemd/unit/tftp' = {
+    if ( OS_VERSION_PARAMS['majorversion'] >= '9' ) {
+        SELF['type'] = 'socket';
+        SELF['state'] = 'enabled';
+        SELF['startstop'] = true;
+    };
+    SELF;
```
ned21, line 22: I believe this will result in the path being set to undef when line 17 is false. I can't remember if having an `else null` works here, or whether that leaves `/software/components/systemd/unit` in the tree which might be a problem on RH6; although maybe RH6 support is OK to drop?

quattor/aii/server/config.pan
```diff
@@ -11,8 +11,18 @@ variable AII_DHCP_DHCPD_CONF_FILE ?= '/etc/dhcpd.conf';
 
 include 'components/aiiserver/config';
 
-# aii-shellfe configuration
+# Enable tftpd (on EL9 only for backward compatibility)
+include 'components/systemd/config';
+'/software/components/systemd/unit/tftp' = {
+    if ( OS_VERSION_PARAMS['majorversion'] >= '9' ) {
```
ned21, line 17: '9' should be 9 instead? I suspect string comparison doesn't work with '10' and higher?

jouvin:
`panlint` and `indentation` errors have been fixed as part of https://github.com/quattor/template-library-core/pull/238. I suggest merging it first and I'll rebase this one to clean the errors.

jrha:
#238 is merged now

quattor/aii/server/config.pan
jouvin: RH6 support has been dropped already in 24.10...

quattor/aii/server/config.pan
jouvin: You are right but for discussable reasons this parameter has been defined as a string (to allow number that may have included a letter or dot, e.g. `9.1`) and the end result is what you are saying. I'm afraid there are several places to fix in the template library and probably a way of doing it would be to implement a function to compare OS versions that would do the proper massaging... I am not clear whether we should already avoid to propagate the problem here or if we should have an issue to fix it properly (and if possible using the same approach) everywhere this test is present in the template library... I'm open!

quattor/aii/server/config.pan
jouvin: But you were perfectly right that the current code doesn't work on EL7/8 as it left a configuration path undefined... This is the reason the `test_templates` was failing! Fixed now, thanks for spotting it!

quattor/aii/server/config.pan
jrha: Would `pkg_compare_version` not do the job?

ned21: I'm not familiar with OS_VERSION_PARAMS and complete agree that overall versioning needs to take things like 8.8 into account. However I would assume that "major version" would only be the first part as an int? i.e. in MS we have similar fields and they are constructed by selecting out the first int before a dot or other character from the overall version. 

Assuming this isn't trivially fixable in the templates then maybe a quick hack here would be (untested):

```
if( to_long(OS_VERSION_PARAMS['majorversion']) >= 9 ) {
```

jouvin: @ned21 I agree with you. The proposed hack would work as major version is currently only something that can be converted to a number. It may not work if at some point we decide to put something that cannot be converted. Thus my proposal to implement a function like `get_os_major_version()` that would return a number after doing whatever is necessary with `OS_VERSION_PARAMS['majorversion']`...

jouvin: Proposal open as an issue:

